PR #4, opened 2018-12-17
MADmin Ui rework

temp666:
Rework the Ui from MADmin

## Description
The Ui is changed to bootstrap and using of datatables is changed.

## Motivation and Context
Better handling of MADmin functions.

## How Has This Been Tested?
Own instance of RD-Map.

## Screenshots (if appropriate):
grafik


cecpk:
Saving mapping editor does not work with that. Could you check this?

PR #8, opened 2018-12-18
A few enhancements for the Madmin Map

sn0opy:
- re-colored known and active spawns to green for better visiblity
- added layer displaying spawns and despawn times
- show actual gym color from the DB
- show further information like spawn times, coords and IDs in popup
- also some code refactoring

Tested and confirmed working with Monocle and RM


AboveColin:
nice!

PR #13, opened 2018-12-20
Stop NULL external_id & IFNULL sponsor & team

georgeherby:
Stop them getting through and breaking sorting when going into JSON


cecpk:
You could merge it by yourself George 😁

sn0opy:
@cecpk yes, but only for OCR people I guess

georgeherby:
Just needed for people that are using the gym_info.json so only OCR i guess

PR #15, opened 2018-12-20
very small change

AboveColin:


muckelba:
python3 or python3.6 is depending on your setup. The other python commands are called python3 in this readme, so i would say keep this at python3 :)

PR #18, opened 2018-12-21
big time fixes to importer

krzee:
rocketmap was kinda fubar, now i have duke's DB and testing is FAR FAR easier. this should work.
I recommend all rm users who already imported run this new version again, but deleting trs_spawn first is now optional, not necessary.


cecpk:
Rdy to go ?

krzee:
ya, now i have duke's db so testing the RM side is about 1000x easier :D

PR #21, opened 2018-12-24
breaking the example config into 2, one for mitm and one for OCR, i n…

krzee:
…eed more eyes please


krzee:
This is not ready to be merged!
Please look over these configs and double check that all the options are where they need to be. I did my best but maybe I did not get it perfect.

krzee:
nevermind! lol

PR #22, opened 2018-12-24
updating example config to separate OCR/MITM mode

krzee:
This is not ready to merge!!!
please take a look and double check the config, im not even 100% positive about some things, like autohatch for example...
but this should get the ball rolling 👌

### DIFF
--- a/configs/config.ini.example
+++ b/configs/config.ini.example
@@ -5,6 +5,10 @@
 # A different config-files can be included directly in processes by "--config-file" or "-cf" and path
 # Value without # are needed for this script
 
+##################################################################
+#           Shared settings - OCR and/or MITM mode               #
+##################################################################
+
 # MySql Settings
 ######################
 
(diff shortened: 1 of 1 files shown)


krzee:
dont just look at the patch, view the whole file please!

krzee:
is that the only problem? i figured there would be more :D

sn0opy:
@krzee could you have another look into this in case this is still a wanted feature?

krzee:
I've been away for a bit, I'm totally unaware of any further config changes. I still think it makes sense, but i wont be home until the end of the month. It's a really easy change to make for anybody who is certain they know what config options are ocr/mitm.

krzee:
The more options that get added the less im able to correctly separate them. If this is a desired enhancement then somebody else should separate them. If not, feel free to just close the ticket

sn0opy:
Closing this for now. Guess we will have another look into it in the upcoming weeks.

PR #23, opened 2018-12-27
update import script to handle weird mariadb versions that dont like …

krzee:
…numeric text strings to be quoted

this is lameness, but it gets the script to work right for @AboveColin , ill check with @eligos85 shortly


eligos85:
now its working

PR #43, opened 2019-01-13
fix overwriting wrong setting

temp666:
fix overwriting wrong setting
if setting name "cityscan" and exists other setting with name "cityscanmon" and you edit "cityscan" then is "cityscanmon" overwriting from "cityscan"


cecpk:
Did you test with all other fields too? 😁

temp666:
yes, tested with all other fields :)

PR #46, opened 2019-01-17
change raid pokemon key to reflect latest changes

sn0opy:
Merge this when latest PogoDroid version has been released to the public.


sn0opy:
not required anymore.

PR #47, opened 2019-01-17
Multiprozessor for calculation_route

Montesuma80:
it is possible to make this script Multiprozessor compatible?


Grennith:
So why exactly is this a pull request?

PR #49, opened 2019-01-19
Dynamically reload settings from mappings.json

michikrug:
This PR adds the functionality to dynamically update the settings for all workers on-the-fly when the mappings.json file has changed.

This enables easy editing of the settings and mappings in MADmin without the need for restarting the whole process to use the new ones.

### DIFF
--- a/start.py
+++ b/start.py
@@ -235,6 +235,10 @@ if __name__ == "__main__":
             except KeyError as e:
                 log.fatal("Could not parse mappings. Please check those. Description: %s" % str(e))
                 sys.exit(1)
+            except RuntimeError as e:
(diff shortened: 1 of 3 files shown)


simonbuehler:
works great, pls merge

sn0opy:
Due to the huge quest/iv update, this PR will break things. Do you mind to have another look into it?

michikrug:
Uhh. Sure, but may take a little while.

michikrug:
@sn0opy should be okay now

michikrug:
Stopping the worker to force it to re-register should solve the problems

michikrug:
Updated to recent changes.

**NOT TESTED!**

PR #50, opened 2019-01-20
Enable display of geofences in MADmin map

michikrug:
As the title states: enables the display of geofences for each area in the MADmin map

**Limitation:** only shows included geofence - NOT the excluded part

_**Code not tested as it is**, since I have a bit different code base already_

**TODO:**
- [x] Adapt latest code changes breaking the PR
- [x] Add support for multiple geofences in one file


sn0opy:
Probably better to have a seperate PR to support the latest changes to the map. Thanks anyway so far!

PR #58, opened 2019-01-31
Proper Quest Tasks

peterbaumert:
Not 100% every task, but the majority is included.

Only the webhook part, if you want this in the database too I leave that to you since I don't want to mess with that :D

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ ocr/weather/*
 configs/config.ini
 *.calc
 configs/geofences/*
+configs/confbak/*
 configs/mappings.json
 .idea/*
 .idea
--- a/utils/questGen.py
+++ b/utils/questGen.py
@@ -97,8 +97,16 @@ def questtask(typeid, condition, target):
                     for ty in pt:
                         arr['type'] += ('or ' if last == cur else '') + pokemonTypes[ty].title() + ('-type ' if last == cur else '-, ')
                         cur += 1
-        if re.search(r"'type': 3", condition):
+        if re.search(r"'type': 3", condition) is not None:
                 arr['wb'] = " with weather boost"
+        match_object = re.search(r"'pokemon_ids': \[([0-9, ]+)\]", condition)
+        if match_object is not None:
+                pt = match_object.group(1).split(', ')
+                last = len(pt)
+                cur = 1
+                if last == 1:
+                    arr['poke'] = pokemonname[pt[0]]
+                text = 'Catch a {poke}.'
         text = "Catch {0} {type}Pokemon{wb}."
     elif typeid == 5:
         text = "Spin {0} Pokestops or Gyms."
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -10,6 +10,7 @@ ocr/weather/*
 configs/config.ini
 *.calc
 configs/geofences/*
+configs/confbak/*
```
georgeherby: I guess this is something you needed? Not for the main repo?

utils/questGen.py
```diff
@@ -77,3 +72,104 @@ def pokemonname(id):
     with open('pokemon.json') as f:
         mondata = json.load(f)
     return mondata[str(int(id))]["name"]
+
+def questtask(typeid, condition, target):
+    with open('utils/quest/pokemonTypes.json') as f:
+        pokemonTypes = json.load(f)
+    with open('utils/quest/items.json') as f:
+        items = json.load(f)
+    throwTypes = {"10":"Nice", "11":"Great", "12":"Excellent", "13":"Curveball"}
```
georgeherby: For consistency of having all this stuff in JSON file, I would expect this to be in a JSON file. Just easier to update if Nia make a change on throws for example

.gitignore
peterbaumert: Yeah my mistake. Gonna remove it. Sorry

utils/questGen.py
peterbaumert: Yeah actually Your right. Dunno why i didnt. Maybe because it were just 4 :D

georgeherby:
@peterbaumert assume you are on Discord? Can you message me, my user the same on Discord as GitHub

utils/questGen.py
```diff
@@ -77,3 +72,104 @@ def pokemonname(id):
...
+    arr = {}
+    arr['0'] = target
+    text = questtype(typeid)
```
evilmoses: You load the text, but replace it in all instances???

Look at type id 5 or 6 for example. Those elifs are not needed?

utils/questGen.py
```diff
@@ -77,3 +72,104 @@ def pokemonname(id):
...
+        if re.search(r"'type': 3", condition) is not None:
+                arr['wb'] = " with weather boost"
+        match_object = re.search(r"'pokemon_ids': \[([0-9, ]+)\]", condition)
+        if match_object is not None:
+                pt = match_object.group(1).split(', ')
+                last = len(pt)
+                cur = 1
+                if last == 1:
+                    arr['poke'] = pokemonname[pt[0]]
+                text = 'Catch a {poke}.'
```
evilmoses: Do you potentially need this too in case there are multiple pokemon?

`else:
        for ty in pt:
        arr['poke'] += ('or ' if last == cur else '') + pokemonname[ty]
         cur += 1            `

utils/questGen.py
peterbaumert: I load the text, so if my code wont edit it the default text will be returned.

peterbaumert:
#62 handles this stuff now, so I close my PR

PR #59, opened 2019-02-01
Fix bugs and clean up MADmin + Make all links relative

michikrug:
_(I know, there is another rework in the making, but still I want to share this with you.)_

Changes:
- Cleaned up and formatted code
- **Fixed wrong request args** access (replace error introduced with basic auth)
- Rewrote some functions that had redundant or even unused code
- **Made all links relative** to use MADmin also on subpath (e.g. with use of a reverse proxy)
- Fixed small display issue with map

### DIFF
--- a/madmin/madmin.py
+++ b/madmin/madmin.py
@@ -136,7 +136,7 @@ def submit_hash():
             copyfile(file, 'www_hash/gym_0_0_' + str(hash) + '.jpg')
             os.remove(file)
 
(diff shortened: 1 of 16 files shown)


michikrug:
@michikrug TODO: redirects are wrong when running in subpath

sn0opy:
See #50. I'd say the same applies to this.

Poke me on Discord, if you like to share the workload.

PR #70, opened 2019-02-08
hopefully fix egg webhooks for PA support

fosJoddie:
whoopsie

### DIFF
--- a/utils/webhookHelper.py
+++ b/utils/webhookHelper.py
@@ -45,7 +45,7 @@ egg_webhook_payload = """[{{
         "gym_id": "{ext_id}",
         "name": "{name_id}",
         "url": "{url}",
-        "pokemon_id": "0",
+        "pokemon_id": 0,
         "sponsor": "{sponsor}",
         "weather": "{weather}",
         "park": "{park}"


evilmoses:
Let me try quickly before we merge

TiMXL73:
PA isn't looking at any value (int or string) to determine if it's an egg or raid, it's just looking if the key `pokemon_id` exists in the payload to differentiate eggs from raids. As such, `pokemon_id` should be removed altogether.

Why was it added though is my question (unused in PA's EggEvent)?

```
elif kind == 'raid' and not message.get('pokemon_id'):
            # RM sends None, M sends 0 for eggs
            return EggEvent(message)
        elif kind == 'raid' and message.get('pokemon_id'):
            # RM/M send Monster ID in raids
            return RaidEvent(message)
```

fosJoddie:
@TiMXL73  did you even read the code you pasted?

```
elif kind == 'raid' and not message.get('pokemon_id'):
            # RM sends None, M sends 0 for eggs
```

will trigger if key pokemon_id is not in message
OR 
if key is there and is the number 0
which is how it has always been with messages from monocle btw

TiMXL73:
Apologies. You are correct and I need sleep.

PR #75, opened 2019-02-12
Detect Weather description change

boby360:
This needs to be enabled in order for weather to be pushed to the database. It is not dependent on OCR thus the need for a description change.


cecpk:
Ok 😜

PR #76, opened 2019-02-12
only update location count if location was actually visited

simonbuehler:


sn0opy:
Thanks, but this is going to be fixed differently, which is why I'd like to close this for now.

PR #81, opened 2019-02-14
Make the log level for websocket library INFO

fosJoddie:
makes websocket logging way less noisy if running in verbose mode and overal


sn0opy:
What kind of log messages does it actually remove?

PR #83, opened 2019-02-15
set center of map to config values

simonbuehler:
center map on home location


sn0opy:
Thanks! What happens, if these values aren't set?

simonbuehler:
hmm, 0,0 i guess ... gonna test it

simonbuehler:
yes, starts in the water, i'll change the zoom factor in that case

simonbuehler:
should be good now

PR #85, opened 2019-02-16
Enable display of geofences in MADmin map

michikrug:
As the title states: enables the display of geofences for each area in the MADmin map

**Important Notice:** to make excluded areas appear, make sure to give the geofence area the same name as the included one (name is used to match polygons)

_**Code not tested as it is**, since I have a bit different code base already_


madmin/templates/map.html
```diff
@@ -283,6 +283,29 @@
 				$this.addLayer("stops with quests", l_quests, false, "quets");
 			});
 		},
+		map_fetch_geofences() {
```
sn0opy: This needs to be called in the `map_fetch_everything()` method to actually fetch the geofences.

sn0opy:
I was also looking into how to display exclusion zones for a polygon and it seems to be quite easy. Only problem is: you have to check whether a certain polygon is within another one to add it to the "same" polygon. Are you interested in checking this beforehand, otherwise I'd merge it.

See https://medium.com/@sumit.arora/what-is-geojson-geojson-basics-visualize-geojson-open-geojson-using-qgis-open-geojson-3432039e336d

madmin/templates/map.html
michikrug: Jep. Somehow slipped through my manual merge

michikrug:
I'll have a look

michikrug:
@sn0opy I reworked the PR to also support excluded areas

sn0opy:
Sorry for the late response and thanks for updating it. I still have one concern that I wasn't yet able to test. What happens to exluded geofences that are (partly) outside of the inclusion geofence? Normally that shouldn't happen, nut well, nobody knows what some users are up to.

michikrug:
Currently, the outside part will be displayed as another included part.
Which is obviously wrong, but this is how the MultiPolygon is working currently.

PR #89, opened 2019-02-17
Webhook URL extension for every type of webhook

temp666:
In my case i need a webhook setting for each type. a few webhooks only need raids and other need raids and quests.
Here is the extension that it is possible to enter your endpoint for each type of webhook.

This needs to test. please give feedback


sn0opy:
I'd prefer to keep just the `webhook_url` setting and like to suggerst a different approach by extending it like this:

Defaults to everything activated:
`webhook_url=http://example.org/foo/bar`

To activate certain options:
`webhook_url=[mon|weather|raid]http://example.org/foo/bar`

With multiple urls:
`webhook_url=[mon|weather]http://example.org/foo/bar,[raid]http://othersite.com/baz`

bbdoc:
What about allowing for an overwrite of this default webhook_url at the area level in mappings.json ? Would be useful if you want to send to different webhook_url depending on the area you are scanning...

sn0opy:
@bbdoc so, you suggest to have an override url per area? That still doesn't really fix the problem that you want to just submit mon for a certain area without sending raids.

bbdoc:
It does if you have this override per area + your idea of
`webhook_url=[mon|weather]http://example.org/foo/bar,[raid]http://othersite.com/baz`
because then you can decide on every area if you want to send to a different webhook for a specific type than the one configured in global config.ini

but that's a nice to have to add even more granularity... don't know if lots of people need this. Wouldn't invest too much time in it before checking there is a real need for it.

temp666:
@sn0opy great Idea. I'll it change to this. That's better than all the keys in the config.

temp666:
It's changed to one `webhook_url` key

sn0opy:
Due to heavy re-work, I'd say we close this for now until #115 has been merged, since it changes the way the webwhook works, thus this PR needs a rebase and rework anyway. I might take over the inclusion of this change if you like.

temp666:
feel free to take over this change.
i need this feature :)

PR #90, opened 2019-02-18
Madmin: comment out routemanager

temp666:
The Routemanager loads all routes when MAD start with -wm.
Routemanager is currently not needed in madmin and he is slows the startup and blows up the logs

### DIFF
--- a/madmin/madmin.py
+++ b/madmin/madmin.py
@@ -50,7 +50,7 @@ def madmin_start(arg_args, arg_db_wrapper):
     db_wrapper = arg_db_wrapper
     mapping_parser = MappingParser(arg_db_wrapper)
     device_mappings = mapping_parser.get_devicemappings()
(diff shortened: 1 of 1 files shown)


sn0opy:
In general, commenting out is not really a good way.

Commenting them out here makes no sense, since the route manager is being initialized in the start.py which then starts madmin

temp666:
Thats true, commenting out is bad :)
When i start Madmin with `sudo python3 start.py -wm` and the Routemanager loads all routes.

See log with Routemanager: https://pastebin.com/ZX3ueZj5
Log without Routemanager: https://pastebin.com/wWAmF2ST

sn0opy:
You're right, `get_routemanagers()` actually initializes everything, not just the initialization of the class.

sn0opy:
Been properly removed in 134bf8b

PR #94, opened 2019-02-19
Disable Monocle Wrapper

cecpk:
:D


bbdoc:
If you only change wiki I'm fine with it ;-) lol

PR #87, opened 2019-02-16
Add more stats

Furtif:

PR #97, opened 2019-02-22
add more stats

Furtif:
fix https://github.com/Map-A-Droid/MAD/pull/87#issuecomment-464856928


sn0opy:
Well, it's still not clear, what this actually fixes or what it actually does 🤔

Furtif:
add more stats form file and create player data for get player name etc.. but part playerdata is no implemented to apk mad Grennith say add this to next

cecpk:
we have to wait for the needed data from pogodroid

Furtif:
@sn0opy  @cecpk  maybe close this

PR #99, opened 2019-02-25
Manual location injection

snooter:
This patch allows you to click on a location in the madmin map to send a worker to scan that spot. The scanner will manually override the routing until the queue is empty, then continue as scheduled.

This is useful for many things, for instance, someone's reported a high quality spawn your scanners haven't found and you want to get it on the map, or you have a hatched egg that the scanner didn't update, or you're traveling to a spot that you don't normally scan but you want to see if there are any raids going on before you leave. It's set up as a database table for maximum flexibility - I have a cron job that's injecting the location of a small town nearby to scan for raids every half hour, one that's not normally scanned. You could also add more frequent scans to important locations with this. 

EDIT: rolled this back, since there could be issues --- There's also another small change bundled in here - the SQL data types in some of the tables don't work with MariaDB, so DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP has been changed to TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP. I'm 99% sure this should work for MySQL as well, but this should be tested to make sure.


sn0opy:
I am running on MariaDB and `DATETIME NOT NULL` always worked for me. What was the actual issuze here?

The idea in general is nice!

peterbaumert:
> I am running on MariaDB and `DATETIME NOT NULL` always worked for me. What was the actual issue here?
> 
> The idea in general is nice!

Not sure what his intention was but something that this helps with is timezones messups :D

"In MySQL 5 and above, TIMESTAMP values are converted from the current time zone to UTC for storage, and converted back from UTC to the current time zone for retrieval. (This occurs only for the TIMESTAMP data type, and not for other types such as DATETIME.)"

sn0opy:
> Not sure what his intention was but something that this helps with is timezones messups :D
> 
> "In MySQL 5 and above, TIMESTAMP values are converted from the current time zone to UTC for storage, and converted back from UTC to the current time zone for retrieval. (This occurs only for the TIMESTAMP data type, and not for other types such as DATETIME.)"

Well, the thing is, that this seems to work just fine and MAD makes sure that all timestamps it is working with are actually UTC. Not sure if this change breaks such behaviour.

snooter:
I went ahead and reverted the TIMESTAMP column changes for this PR, to avoid complicating things. Is there a different way to address the MariaDB problem? It's an issue with versions before 10.0:

MariaDB [pgm]>  Create table if not exists trshash (
    ->                   hashid MEDIUMINT NOT NULL AUTO_INCREMENT,
    ->                   hash VARCHAR(255) NOT NULL,
    ->                   type VARCHAR(10) NOT NULL,
    ->                   id VARCHAR(255) NOT NULL,
    ->                   count INT(10) NOT NULL DEFAULT 1,
    ->                   modify DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
    ->                   PRIMARY KEY (hashid));
ERROR 1067 (42000): Invalid default value for 'modify'

Others have had the same issue, and this solution has worked as well: https://discordapp.com/channels/465247740553592832/467318193141317659/533406155938988032

sn0opy:
Well, that's because the query is actually wrong. Change the default value to `current_timestamp()` (notices the braces) and it will work.

sn0opy:
> I went ahead and reverted the TIMESTAMP column changes for this PR, to avoid complicating things. Is there a different way to address the MariaDB problem? It's an issue with versions before 10.0:
> 
> ...
> ERROR 1067 (42000): Invalid default value for 'modify'
> 
> Others have had the same issue, and this solution has worked as well: https://discordapp.com/channels/465247740553592832/467318193141317659/533406155938988032

Well, that's because the query is actually wrong. Change the default value to current_timestamp() (notices the braces) and it will work.

snooter:
The parens don't help for MariaDB 5.5, same error: 

MariaDB [pgm]> Create table if not exists trshash (
    ->                 hashid MEDIUMINT NOT NULL AUTO_INCREMENT,
    ->                  hash VARCHAR(255) NOT NULL,
    ->              type VARCHAR(10) NOT NULL,
    ->                  id VARCHAR(255) NOT NULL,
    ->                  count INT(10) NOT NULL DEFAULT 1,
    ->                    modify DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP(),
    ->               PRIMARY KEY (hashid));
ERROR 1067 (42000): Invalid default value for 'modify'

MariaDB forked from MySQL before it added support for DEFAULT CURRENT_TIMESTAMP in DATETIME columns in MySQL 5.6.5, so it took them awhile to catch up with that feature. Upgrading MariaDB would be a pain because I have a ton of other stuff running reliably on it, and 5.5 is still heavily used by others with LTS support until 2020, so it'd be great to find a solution for this (though probably not in this thread).

sn0opy:
image

🤷‍♂️

sn0opy:
Would you like to add an extra config parameter (not really sure if this should be enabled by default) that "disables" the actual injected location check? Something like `enable_injected_location_check`. This can then be used to disable the DB lookup since it could raise DB load if there's an extra lookup for every new location while having multiple workers.

With that little extra I'd be happy to merge this.

snooter:
How does this work? I created a central repository object to handle periodic refreshes of the location injection requests from the database (set to query the database at most every 5 seconds, no matter how many workers there are). 

As far as choosing the specific worker instead of the mode, I set it up this way initially, but changed it to this when I realized I want to say "send one of the raid scanners to investigate this spot", rather than trying to remember which workers are currently handling raids and choosing one of them.

sn0opy:
@cecpk build his own version right into the phone control update. I guess we can close this for now. We might come back to this later, since having some sort of queue would be quite useful.

PR #107, opened 2019-02-28
Add only_routes mode

crhbetz:
Quick implementation of `-or` / `--only_routes` mode, which will calculate missing route files while being able to keep your working instance running alongside it, as requested on Discord every couple of days.

Usually, calculating a complicated route at the beginning of the program causes a downtime. Running with -or separately will avoid this and shouldn't collide with the main instance in any way.

It's implemented as just a `sys.exit(0)` at the appropriate location in `start.py` when `-or` is set.

### DIFF
--- a/start.py
+++ b/start.py
@@ -243,7 +243,7 @@ if __name__ == "__main__":
                   "-oo     ---- start OCR analysis of screenshots\nExiting")
         sys.exit(1)
 
-    if args.only_scan:
+    if args.only_scan or args.only_routes:
         
         filename = os.path.join('configs', 'mappings.json')
         if not os.path.exists(filename):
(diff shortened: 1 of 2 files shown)


sn0opy:
You're actually required to run this with `python start.py -os -or` to actually work. You probably want to add the parameter to https://github.com/Map-A-Droid/MAD/blob/090e48cbc809a47539f1f3df5d0cb07fa1e6b665/start.py#L239 as well.

crhbetz:
> You probably want to add the parameter to
> 
> https://github.com/Map-A-Droid/MAD/blob/090e48cbc809a47539f1f3df5d0cb07fa1e6b665/start.py#L239
> as well.

I think after adding the parameter here, it should work without `-os`. Thanks for reviewing!

PR #109, opened 2019-03-01
Add is_in_battle and EX Gyms

zaeshu:
RocketMap:
1. is_in_battle is updated in the DB for Gyms.
2. is_ex_raid_eligible is updated from raids and gyms - Webhook is sent.
3. I removed the weather_boosted_value because the raid_mon is only updated after encountering
This requires a DB modification:
ALTER TABLE 'gym'
ADD 'park' tinyint(1) NULL;

### DIFF
--- a/db/rmWrapper.py
+++ b/db/rmWrapper.py
@@ -170,11 +170,12 @@ class RmWrapper(DbWrapperBase):
                 "UPDATE raid "
                 "SET level = %s, spawn = FROM_UNIXTIME(%s), start = %s, end = %s, "
                 "pokemon_id = %s, last_scanned = FROM_UNIXTIME(%s), cp = %s, "
-                "move_1 = %s, move_2 = %s "
+                "move_1 = %s, move_2 = %s, weather_boosted_condition = %s "
                 "WHERE gym_id = %s"
             )
             vals = (
-                lvl, now_timestamp, start_db, end_db, pkm, int(time.time()), '999', '1', '1', gym
+                lvl, now_timestamp, start_db, end_db, pkm, int(time.time()), '999', '1', '1', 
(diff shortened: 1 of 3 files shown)


db/rmWrapper.py
```diff
@@ -445,7 +447,7 @@ def get_near_gyms(self, lat, lng, hash, raid_no, dist, unique_hash="123"):
             ") "
             "AS distance, latitude, longitude, name, description, url "
             "FROM gym "
-            "LEFT JOIN gymdetails ON gym.gym_id = gymdetails.gym_id "
+            "LEFT JOIN gym_details ON gym.gym_id = gym_details.gym_id "
```
TiMXL73: Shouldn't this be gymdetails and not gym_details for RM's database?

```suggestion
            "LEFT JOIN gymdetails ON gym.gym_id = gymdetails.gym_id "
```

zaeshu: That isn't part of my commit. Not sure why that's in there. Snoopy fixed it last commit

db/rmWrapper.py
```diff
@@ -170,11 +170,12 @@ def submit_raid(self, gym, pkm, lvl, start, end, type, raid_no, capture_time, un
                 "UPDATE raid "
                 "SET level = %s, spawn = FROM_UNIXTIME(%s), start = %s, end = %s, "
                 "pokemon_id = %s, last_scanned = FROM_UNIXTIME(%s), cp = %s, "
-                "move_1 = %s, move_2 = %s "
```
sn0opy: You removed this and a few more spaces which are actually required and things will break without them.

zaeshu:
Closing due to rework of webhooks

PR #112, opened 2019-03-04
Do not try to delete Mystery Box when clearing box

crhbetz:
There's probably not a lot of worker accounts with a Mystery Box, but those fail to clear bag space without this little addition.


cecpk:
rly dont understand this ... but ok 🗡 
DONT CONNECT YOUR SWITCH TO A MAD WORKER !!!!!

PR #113, opened 2019-03-04
Fix Quest Webhook

michikrug:
Currently, the quest webhook sends some data that is mostly not well processable by e.g. BookOfQuests.

After some discussions with the guy from BoQ, we found a solution to make the current quest webhook messages compatible.

Changes:
- also store and deliver `quest_template`
- include all non-string values correctly (int, float, json)
- fix the condition to be valid json

__As usual from my side: untested as it is__

_(I also have some changes in my own code base that enables compatibility with PokeAlarm dev, but this is not included here.)_

### DIFF
--- a/db/dbWrapperBase.py
+++ b/db/dbWrapperBase.py
@@ -854,6 +854,7 @@ class DbWrapperBase(ABC):
         if 'challenge_quest' not in map_proto:
             return False
         quest_type = map_proto['challenge_quest']['quest'].get("quest_type", None)
+        quest_template = map_proto['challenge_quest']['quest'].get("template_id", None)
         if map_proto['challenge_quest']['quest'].get("quest_rewards", None):
             rewardtype = map_proto['challenge_quest']['quest']['quest_rewards'][0].get("type", None)
             reward = map_proto['challenge_quest']['quest'].get("quest_rewards", None)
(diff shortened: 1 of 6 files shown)


sn0opy:
I'd like to wait a bit longer until #115 has been merged. If you're intersted, you can rebase this onto the https://github.com/Map-A-Droid/MAD/tree/webhook-worker branch and make it work with it, so both things will get into master at the same time.,

sn0opy:
I'll have another look into this tomorrow. Probably going to merge this into webhook-worker instead which will add the ability to send a payload only to certain webhook endpoints.

sn0opy:
Just to clarify why this has not been merged yet. Due to the new webhook worker, data needs to be inside of the DB to be sent off to the webhook endpoint. 

This PR requires some extra fields which MAD does not yet save to the DB. So, we have to add these fields to the DB first. After that, we can implement this into the webhook worker.

cecpk:
@michikrug i saw one new field that you already added in a version update for mad. So we ""just"" need the integration into new webhook system. @sn0opy how could we add a switch for the Endpoints? Do we need more then one Endpoint? I dont see any support for quests in other systems like PA in near future. Based on the URL is not good.
Whats the unique key for sending a quest? The pokestop id - external id?

sn0opy:
> @sn0opy how could we add a switch for the Endpoints? Do we need more then one Endpoint? I dont see any support for quests in other systems like PA in near future. Based on the URL is not good.

I'd totally go with the BoQ standard for now and we will try to come up with a different idea for the future when other receives are able to receive quests.

> Whats the unique key for sending a quest? The pokestop id - external id?
I'd say the pokestop id + quest id, since there can be different quests over the day due to events and you probably don't want to re-scan after the event is over.

michikrug:
> I dont see any support for quests in other systems like PA in near future.

I use PA dev for quests (slightly modified)

michikrug:
> This PR requires some extra fields which MAD does not yet save to the DB. So, we have to add these fields to the DB first. After that, we can implement this into the webhook worker.

As I was in close contact with the BoQ guy, he said that the new field is currently not required but may be useful for future quest types. Therefore, and to be compatible with the other RDM tool's data, I added this to MAD too.

cecpk:
and you use the same data for your mod. PA?

michikrug:
Yes, with the addition of sending our generated quest text too. Since PA has no logic to construct the quest task and reward.

PR #115, opened 2019-03-06
WIP: Add webhook worker

sn0opy:
This PR is meant to enhance the general behaviour MAD sends data to webhook endpoints. The current behaviour sends out multiple messages for A. every GMO received and B. for every element that is contained in it. Thus one single GMO could result in several dozens of single requests being performed.

The webhook worker is a seperately running thread to keep everything async. Every 10 seconds it pulls modified data from the DB and aggregates everything into one single payload which is then sent off to the webhook endpoint.

The current behaviour of this branch does NOT consider duplicates. If the timestamp within the DB changed, it will be sent to the webhook endpoint eventhough it was sent with the previous iteration already.

Thanks to @temp666's PR #128, this branch also includes type-specific webhook urls. See his PR for further information.

**Testing:**
To test this PR, you can run `git checkout webhook-worker` within your MAD directory. It will then switch over to the branch which is based off of `master`. 

**Status (2019-03-11 11:15pm UTC):**
RM ~should~ already works fine ~for most parts. See todo list for more details.~

~Monocle is basically untested, but could already work with weather, mon and raids. It won't send any gym updates, since the code has been removed already.~ Confirmed working

**Todo:**
- [x] RocketMap: raids
- [x] RocketMap: weather
- [x] RocketMap: gyms
- [x] RocketMap: mon
- [x] Monocle: raids
- [x] Monocle: weather
- [x] Monocle: mon
- [x] Monocle: gyms
- [x] Remove special methods from db wrappers (like mon level calculation)


webhook/webhookworker.py
```diff
@@ -0,0 +1,263 @@
...
+
+        for raid in raid_data:
+            # skip ex raid mon if disabled
+            if (
+                not self._args.webhook_send_exraid
+                and raid.pokemon_id == self.EXRAID_MON_ID
+            ):
+                pass
+
+            raid_payload = {
```
fosJoddie: There's a few things missing here that isn't kept in RMs database but is both used in PA and other raid webhook receivers:
`park` - is a flag for if the gym is - in GMO its the `gym_details.is_ex_raid_eligible` data - this field is not saved in DB
`is_exclusive` flags if the raid is an ex raid and is "by invitation only" PA does not support this, but other tools do, and so should PA. It can be used both to filter out sending of ex raids, or to filter alerts by only ex raids (for whoever wants that). This flag is in the `gym_details.raid_info.is_exclusive` part of GMO and is not saved to the RM raid table

fosJoddie:
A solution to any piece of webhook data that will get lost because it isn't ain't in the DB schema might be to simply add `trs_gym` and `trs_raid` tables that keep surplus data for us. 

While I can agree with the approach of asking the db "what's new" and send webhooks of that, I don't like that we remove any data the db isn't fit to send. The beauty of having the pure protos available when sending the webhooks was the way you could just send stuff that your db backend maybe didn't even save. 

Trying to look through the code, I'm not sure exactly what, if any, fields we are sending today that will be removed with this PR though

PR #116, opened 2019-03-10
Switch section Devices with Areas.

JabLuszko:
After initial setup I found that if I needed to change something it was mostly in Devices section - it was kinda annoying to scroll past all the configured Areas to get to Devices. Devices first, Areas second, Auth last.


cecpk:
ok for me @sn0opy

PR #123, opened 2019-03-11
Fix problems with timestamps being in local time

Jaap-Jan:
Timestamps were in local time, not UTC, thus not UNIX timestamps.


sn0opy:
I might be wrong on this one, but I think they're in local time because OCR times are most like in local time. @cecpk what do you say?

Jaap-Jan:
In `detectRaidEndtimer()` and `detectRaidTime()`, the timestamp is used for comparison, so that should work regardless of whether the timestamps are local or UTC, as long as the time zone for the time zones match. Note that the code in `detectRaidTime()` that uses the timestamp is commented out at the moment, but it should be correct.

In `getEndTime()`, the end time of an already hatched egg is calculated. The detected time is the time in minutes and seconds until the end of the raid. So this should also work for local time and UTC time.

The only place where local time is actually used, is `getHatchTime()`. Here, a local time is created, with the time set to midnight. Then the timestamp is converted to a UNIX timestamp with the call to `datetime.timestamp()` and then detected time is added.

So in `getHatchTime()` it is assumed that the server's time zone matches the phone's time zone. I think this isn't a weird assumption, but maybe it's better to allow to set phone's time zone in the configuration.

cecpk:
good question - could be ...
could you check RM too @Jaap-Jan ?

PR #125, opened 2019-03-12
submit_quest_proto Exception handling

bbdoc:
Adding a try/except block around submitting of quest proto and error log in case of exception.

Adding some log warning when no fort_id is set or no challenge_quest in proto.

This will allow for a better understanding of why some quests are not being inserted into the DB.

### DIFF
--- a/db/dbWrapperBase.py
+++ b/db/dbWrapperBase.py
@@ -848,43 +848,45 @@ class DbWrapperBase(ABC):
 
     def submit_quest_proto(self, map_proto):
         log.debug("{DbWrapperBase::submit_quest_proto} called")
-        fort_id = map_proto.get("fort_id", None)
-        if fort_id is None:
-            return False
-        if 'challenge_quest' not in map_proto:
-            return False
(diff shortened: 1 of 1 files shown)


sn0opy:
Is this actually require to be PR'd? I'd prefer to have a solution for the actual problem instead of having this within the code that is not actually needed by the whole userbase.

Grennith:
What exception is this supposed to handle? Catchall isn't really nice here. It could be key/values missing in the passed dicts, some uncaught exception in the SQL related stuff or god knows

bbdoc:
I'm not sure what the actual problem is, might be pogodroid based. Still I think it's useful to have some visibility in logs when a quests doesn't follow the normal path. With the 2 log.warning, at least we already catch it if linked to fort_id or challenge_quest missing in proto... exception is for anything else. I agree we could add more granularity, but at the moment we don't even catch it at all.

I'm pretty sure lots of users have it at the moment but don't even realize.

Once the actual issue is found and solved, this exception block won't even be entered anymore so doesn't really matter to have it added.

cecpk:
think the problem was a other one ...

PR #128, opened 2019-03-13
Webhook URL extension for every type of webhook

temp666:
A few webhooks only need raids and other need raids and mons.
Here is the extension that it is possible to enter your endpoint for each type of webhook.

Defaults to everything activated:
`webhook_url=http://example.org/foo/bar`

To activate certain options:
`webhook_url=[mon|weather|raid]http://example.org/foo/bar`

With multiple urls:
`webhook_url=[mon|weather]http://example.org/foo/bar,[raid]http://othersite.com/baz`

### DIFF
--- a/utils/walkerArgs.py
+++ b/utils/walkerArgs.py
@@ -152,7 +152,7 @@ def parseArgs():
     parser.add_argument('-wh', '--webhook', action='store_true', default=False,
                         help='Activate webhook support')
     parser.add_argument('-whurl', '--webhook_url', default='',
-                        help='URL endpoint/s for webhooks (seperated by commas) - urls have to start with http*')
+                        help='URL endpoint/s for webhooks (seperated by commas) with [<type>] for restriction like [mon|weather|raid]http://example.org/foo/bar - urls have to start with http*')
     parser.add_argument('-pwh', '--pokemon_webhook', action='store_true', default=False,
                         help='Activate pokemon webhook support')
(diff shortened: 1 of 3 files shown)


sn0opy:
Looks good to me. Thanks!

utils/walkerArgs.py
```diff
@@ -152,7 +152,7 @@ def parseArgs():
     parser.add_argument('-wh', '--webhook', action='store_true', default=False,
                         help='Activate webhook support')
     parser.add_argument('-whurl', '--webhook_url', default='',
```
fosJoddie: I think you can make multiple webhook urls easier to setup if you add ` action='append'` to the parser argument definition.

Then you can simply do: 

```
webhook_url: [mon|weather|raid]http://example.org/foo/bar 
webhook_url: [raid]http://otherUrl.org/foo/bar 
```
for a less cluttered way of configuration.

sn0opy: I like the idea, thing is: what's easier to explain? The extra line or the comma seperated list?

fosJoddie: I'd say it's easier to just add lines, the users who need more than one would probably be techy enough to figure it out any which way

sn0opy: Okay, aggreed. @temp666 if you like, feel free to PR this as well, otherweise I'm going to work on it next week.

PR #130, opened 2019-03-14
Update RM DB with Gyms in Battle & EX Eligible Gyms

zaeshu:
This will update RocketMap database columns is_in_battle and park under table gym.

### DIFF
--- a/db/rmWrapper.py
+++ b/db/rmWrapper.py
@@ -846,12 +846,13 @@ class RmWrapper(DbWrapperBase):
 
         query_gym = (
             "INSERT INTO gym (gym_id, team_id, guard_pokemon_id, slots_available, enabled, latitude, longitude, "
-            "total_cp, is_in_battle, last_modified, last_scanned) "
-            "VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s) "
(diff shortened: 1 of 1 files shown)


cecpk:
do we have this fields in RM DB? 🍡

temp666:
yes, with fresh db.
But for exists db is no migration in OSM-RocketMap.

sn0opy:
May I suggest one change please? I'd rather have a seperate field to be compatible with Monocle. 

Instead of using `park`, we should actually use `is_ex_raid_eligible`, so we can add the same field to the Monocle db as well, since Monocle's fort's `park` collumn works differently.

krzee:
monocle park just needs to be set to 1 and then park_id could be set to anything non null/0

sn0opy:
> to be supported by pmsf monocle park just needs to be set to 1 and then park_id could be set to anything non null/0

it could, but that's not the way Monocle works. The `park` field is a varchar that holds the park's type.

krzee:
i dont think monocle is coming back, only thing keeping the format alive is MAD... as long as its valid to pmsf while looking at the monocle db id think mad is free to do as it likes here :)

sn0opy:
We now have an extra field in RM and Monocle which handles this.

PR #138, opened 2019-03-22
fix jquery blockUI on https

crhbetz:
Otherwise gets blocked as mixed content when accessing through https.


evilmoses:
Thanks. Working fine for me.

PR #139, opened 2019-03-22
Allow the user to show all quests at once with the "All" selection

fosJoddie:
Small QOL change for big boys with lots of quests and no time for paging


cecpk:
Hack the system!!

PR #140, opened 2019-03-22
Madmin quests: Make the reward picture and pokestop picture hide first

fosJoddie:
When we run out of space on phones, hide the Photos first from the table so that the quest task and reward can still be seen.
Sets autowidth to false since it seems to handle better on phones.

Needs testing, plz no auto merge


cecpk:
nice on mobile! merged to dev

PR #141, opened 2019-03-25
Fix a typo, I guess.

JabLuszko:
It's actually called webhook_submit_exraids in config.ini.example so I assume we using this name. If not just fix in config to the one used in webhooks.

### DIFF
--- a/webhook/webhookworker.py
+++ b/webhook/webhookworker.py
@@ -127,7 +127,7 @@ class WebhookWorker:
         for raid in raid_data:
             # skip ex raid mon if disabled
             is_exclusive = raid["is_exclusive"] is not None and raid["is_exclusive"] != 0
-            if not self.__args.webhook_send_exraids and is_exclusive:
+            if not self.__args.webhook_submit_exraids and is_exclusive:
                 continue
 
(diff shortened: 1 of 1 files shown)


JabLuszko:
and tbh I am not sure if that exraids should be global option, not 'per webhook' in square brackets - but then you need to decide what is 'default' - do MAD sends it as webhook and you name option like 'disable_exraids' or you make it off by default and people need to read docs and enable that.

sn0opy:
> and tbh I am not sure if that exraids should be global option, not 'per webhook' in square brackets - but then you need to decide what is 'default' - do MAD sends it as webhook and you name option like 'disable_exraids' or you make it off by default and people need to read docs and enable that.

Sending ex raids was always off by default. We never did. Now it's a global option to enable it, since it's still off by default.

JabLuszko:
> Sending ex raids was always off by default. We never did. Now it's a global option to enable it, since it's still off by default.

No, it wasn't off by default. I got ex-raid eggs as long as I am running MAD (so like 3 months?). I didn't got mons hatching, but it wasn't MAD doing, it's Niantic configuration thing.

PR #145, opened 2019-03-26
Added encounter device sync

masterxq:


masterxq:
added 1 hour extra despawn time for solving issues with the unknown spawn times

masterxq:
Same encounter id occurs multiply times in monocle (sightings), it should be checked if this is correct behavior

masterxq:
Checked as fine, should be complete.

cecpk:
double - merged into dev

PR #152, opened 2019-04-03
mysql-connector==2.1.6 has been removed from PyPi

Azelphur:
This causes new installs to fail when pip installing the requirements, 2.1.7 is available, so this changes the requirements.txt file to use 2.1.7


fosJoddie:
I see the project has been replaced with - https://pypi.org/project/mysql-connector-python

is there any blocing issues preventing us to change the requirement to that one?

Azelphur:
I installed mysql-connector-python and gave it a quick test, things do seem to be working with mysql-connector-python, at least for my 2 minute test run :)

PR #154, opened 2019-04-04
Update config.ini.example

JabLuszko:
Deleting those (for ocr only mode) misleading info. We need DB for everything. [Another 1 line commit, straight to master! ;)]


cecpk:
Pls add a hint for not using none as ip ;)

JabLuszko:
More like set to localhost for default/empty :>. I will check later then.

PR #176, opened 2019-04-17
Validate walkervalue.

JabLuszko:
We force two digits for minutes (just like the description by cec states ;)), so people won't end with 1:1, it must be 1:10 or 1:01. Whitespaces between hyphen allowed. This regexp validate for non-existing hours, so no more 25:49 or other typos.

No jquery btw. :-D


cecpk:
so just 1 is valid - f.e. 1 round? :D

JabLuszko:
This one should be valid for most of the stuff, but check and maybe merge that into dev first.
Welcome to little spaghetti code, but this way we could reuse those vars I guess.

PR #177, opened 2019-04-18
Dispatch workers to location by tapping on map

snooter:
Building on the new phonecontrol sendGps feature by allowing a location injection by tapping on the madmin map. Also reworked the feature slightly to eliminate the need to pass the adb parameter to send_gps(), since it's easily accessed through device_mappings.


cecpk:
Why did you kill den adb parm? Not needed for geo fix - but for the screenshot after fix 😜 pls edit that. Good idea

snooter:
@cecpk Sure thing, how's this?

sn0opy:
I really like the idea. Just thinking about two more things:

- not sure about a single click triggering the modal. I have something like that Google Maps icon in mind. The one you can drag to the position to trigger Streetview at a certain place

- setting a sleep time is a permanent thing. At least until MAD is being restarted. It'd be nice to have this reset after the injection happened. One idea to make this possible: an extra thread with a sleep time that will reset the sleep time back to the config value.

snooter:
@sn0opy Google Maps already has a behavior assigned to a single click on the map (showing the address and thumbnail image of that spot). In this map, a single click doesn't have any other purpose, and is a lot easier and intuitive from a user standpoint than dragging an icon.

cecpk:
i want drag 🗡

sn0opy:
What I actually meant is that, a single click can quite accidental. But I guess, everything else is just QOL. Will merge this in a bit.

PR #180, opened 2019-04-22
madmin quests: add pokestop lat / lon and link it to OSM

sprintnudel:


sn0opy:
Why OSM if everything else is linking to Google Maps?

sprintnudel:
You guys use OSM everywhere, no? Generally I prefer OSM because it has more detail and usually is mor eup to date compared to Google maps.

sn0opy:
Uhm, no. We don't. Every link on the madmin map points to Google Maps, so I'd say we stick to Gmaps here too for now. Feel free to add an extra option to let instance owners decide which one they want to use.

sprintnudel:
Yeah in that case it makes sense. Will open another PR.

PR #182, opened 2019-04-22
add missing parameter

nhumber:
add missing parameter to rm style db , monocle db style already updates this field


sn0opy:
The problem with this parameter and why it wasn't implemented is basically because we can't keep it updated every second/minute, thus the data is long outdated before we are able to update it the next time.

sn0opy:
Closing this for now for the given reason.

PR #185, opened 2019-04-23
Updated addresses.json for 0.141.0

sn0opy:
Merge this if you either want to use 0.141.0 or when it has been forced.


sn0opy:
0.141.1 has already been released

PR #187, opened 2019-04-24
WIP Add support for Poracle compatible quest webhooks

fosJoddie:
Adds a new optional argument `quest_webhook_flavor` that can be either default or poracle. 

If set to poracle, quest webhooks will be sent in a format that is compatible with PoracleJS :tm:

Includes a new try/catch block for quest gen in case something blows up.

Needs extensive testing by people using Poracle to ensure all quest conditions and rewards come through correctly.


fosJoddie:
Known issues: 
- some quests fail to be sent because of some bad programming - need logs and info about which quests, but current error log is lackluster

PR #189, opened 2019-04-24
Implement a quicker but less effective way to calculate routes

fosJoddie:
Make it possible to sacrifice route optimization in favor of much faster route calculcations

new setting: 
`route_calc_algorithm`

default value is `optimized` - alternate value is `quick` to use the alternative calculations

### DIFF
--- a/configs/mappings_example.json
+++ b/configs/mappings_example.json
@@ -104,7 +104,8 @@
                 "post_walk_delay": 0,
                 "reboot": false,
                 "restart_pogo": 0,
-                "walk_after_teleport_distance": 1
+                "walk_after_teleport_distance": 1,
+                "route_calc_algorithm": "quick"
             },
             "walker": "test123"
         },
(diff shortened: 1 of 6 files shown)


fosJoddie:
Benchmark:

Calculating a quest route of almost 300 stops:
Quick: 305 ms
image

Optimized: 119253 ms (almost 2 minutes)
image


As you can see, the quick route is visibly "worse"

bbdoc:
Love it... mainly for quests as mine can take up to 48h to recalculate... so in the end I'm working with non optimized routes as well as new stops are not being visited. Will test ;-)

fosJoddie:
Rebased against dev and moved the new setting into device setting

krzee:
this is useful when processing a large route on a raspberry pi <3
thank you

fosJoddie:
rebased against latest dev

korentzari:
Add option on zone configuration and not devices ?

cecpk:
Merged in dev - edit: setting on area not on device

PR #192, opened 2019-04-28
urllib3 version incompatibility fix

boby360:
Using urllib3>=1.24.2 would update to 1.25.1 which isn't compatible with the current requests version used by this file. Forcing version 1.24.2 works as it should.


sn0opy:
Just pushed a cleaned-up version of the requirements.txt wich fixes this and prevents similar issues in the future. Thanks anyway.

PR #198, opened 2019-04-30
Speedy quests

fosJoddie:
Speed up quest handling a bit with a few subtle changes

* When clearing inventory, don't crash and burn if tesseract has a bad day and returns bad unicode
* If the setting for walk after teleport is on, remove the time spent walking from the waiting time between quests
* If we aren't receiving quest after open pokestop, retry this 3 times instead of 2 - and if already retried twice, let's check DB if maybe we actually got it before the timestamp set (sometimes happens)


cecpk:
Merged

PR #200, opened 2019-05-05
Fix lookup of existing spawndef for updating spawnpoints

inSilex:
Currently the dictionary provided by getspawndef has strings for keys.
However, when the function is used - while attempting to update spawnpoints - we treat the spawnpoint id as an int. This causes the lookup to fail, and we always overwrite spawndef instead of updating it.
This converts the keys in getspawndef to integers.


sn0opy:
Hey, I've just pushed your fix directly to dev.

PR #202, opened 2019-05-07
General MADmin (map) rework

sn0opy:
The map becomes pretty slow the more elements one has. This PR is meant to fix this with the following changes

**RM ONLY FOR NOW**

* new setting to choose maptiles
* beautified tooltips 
* re-enable double-click zoom
* move routes and geofences to their own sidebar tab
* routes now show their type
* only load elements, if they're enabled
* only load enabled elements within viewport
* automatically update elements every n seconds
* general improvements, by moving more elements into the data storage and generate popups dynamically
* fix a few smaller bugs like a stuck map when showing and hiding quests

**Screenshots:** 
image

image

image

image

image

**Known issues:**
* [x] only supports RM for now 9fa9a95
* [x] changing coordinate radius settings is broken 2f32853
* [x] spawn popups are a bit off since they're using the gym popup style 7988c46
* [x] overlapping text in quest popups 181e183

This PR is based off of `dev`. To use it, run `git checkout madmin-rework`


fosJoddie:
Initial testing says 👌 

I like the new colors of the routes and the snappiness of loading just what is in view!

PR #203, opened 2019-05-07
Force ditto as pokemon_id if ditto.

JabLuszko:
Use only if you want to see ditto in madmin quests/webhooks ASAP.


Christian-X:
Preferred way would be to export the is_hidden_ditto flag to the webhook as well. This way quests webhook users can see the pokemon that is visible first AND have the information that it is a hidden_ditto...

sn0opy:
As mentioned by @Christian-X this should be handled a bit differently. Same should apply to the internals to make sure we know the actualy mon and that this is a Ditto. Going to close this for now, since it's not relevant anymore.

We might come back to this when it gets relevant again.

PR #218, opened 2019-05-22
update_stopsgyms.sh now copies pokestop name/image info to gyms that …

krzee:
…used to be pokestops


krzee:
Could a Monocle user please test this?

PR #220, opened 2019-05-26
Recent shinies on statistics page

JoniValles:
*Only RM
**Need to create column "is_shiny" on Pokemon table.


sn0opy:
Since this only supports RM, has no proper way to add the missing collumn and an alternative is already in the internal work I'm going to close this one for now. Thaks anyway for your work!

PR #222, opened 2019-05-27
Added lure support and pokestop webhook

JimBim89:
Added support for webhook type "pokestop" which is supported by PA for example.
Also modified rm wrapper to pick up information to fill the needed lure data.

PS: for now it will always assume a non special lure duration.


webhook/webhookworker.py
```diff
@@ -458,6 +458,32 @@ def __prepare_gyms_data(self, gym_data):
...
 
+    def __prepare_pokestops_data(self, pokestop_data):
+        ret = []
+
+        for pokestop in pokestop_data:
+            if self.__is_in_excluded_area([pokestop["latitude"], pokestop["longitude"]]):
+                continue
+
+            pokestop_payload = {
+                "pokestop_id": pokestop["pokestop_id"],
```
fosJoddie: I know PA doesn't (currently) support it, but add pokesotp name and image url here as well.

They would add so much worth.

sn0opy:
Already modified this PR to run against dev branch. Please change your code locally to run against the dev branch too and fix those issues appearing in this PR.

webhook/webhookworker.py
JimBim89: Its already included, I also opened a PR for PA to support it.

db/rmWrapper.py
```diff
@@ -1214,6 +1221,30 @@ def quests_from_db(self, neLat=None, neLon=None, swLat=None, swLon=None, oNeLat=
                 'task': quest_task, 'quest_template': quest_template})
 
         return questinfo
+        
+    def get_pokestops_changed_since(self, timestamp):
+        query = (
+            "SELECT pokestop_id, latitude, longitude, lure_expiration, name, image, active_fort_modifier "
+            "FROM pokestop "
+            "WHERE DATEDIFF(lure_expiration, '1970-01-01 00:00:00') > 0 AND last_updated >= %s"
```
sn0opy: This makes no sense.`lure_expiration` is always "1970-01-01 00:00:00" causing this method retuning dozens to hundrends of stops every 10s, since they all become updated.

Since there's actually no reason to post just stops to a webhook, I'd rather say we change this to something like `get_lures_changed_since()` just to retrieve lure modules. Thus we can add proper checks (see main thread in this PR)

webhook/webhookworker.py
fosJoddie: 🤦‍♂ yeah I see that you did that...

db/rmWrapper.py
JimBim89: It makes sense since it will only compare recently scanned entries and looks for a non default lure_expiration (which is a real date if a lure is active). Sure this can be tweaked but its not as bad?

sn0opy:
My general thoughts on lure modules:

- `last_modified` - is giving us a timestamp of the time a lure module has been probably been installed.
- `last_seen_lure` -  (new) to give us a timestamp when we've seen the lure module first. helpful for CDs/events
- `last_updated` - the timestamp a stop has been scanned and updated the last time.

In general, we use 30 min as a default time. I prefer to keep handling of the actual duration on the webhook receiver's end / frontends. MAD should just return `last_updated` timestamp imho.

Beside the value of `active_fort_modifier` we can also check if `last_seen_lure == last_updated`. If true there's definetely a lure module active. 

Pseudo-code for possible CD detection: `if (last_seen_lure == last_updated) && last_updated > last_modified` -> CD/event

Facts:
* we know, when module has been installed
* we know, if there's still an active module the last time we scanned a stop

Problems:
* we don't know the actual duration of a lure module. For Events and CDs, you have to set the lure duration manually

JimBim89:
Yes, I can understand your desire to also support special lure duration times.

Sure after half an hour over the modified time and with still having a lure active we can assume a special event duration time, but without having the exact extra duration and depending on the endpoint for the data this would still not be that useful. 

A message endpoint (that posts to discord, fb, telegram etc) needs an endtime when you first see the lure, and I rather would not have multiple messages for one lure, I would just assume it hasn't ended when I know that we currently have a longer duration.

For the live map on the other hand I would like to have them displayed as active after the first half hour, cause I didn't get to see the all endtimes and its just not visible anymore.

TLDR: Without adding a special event duration time with start and end dates in the config or acquiring the exact raw endtime from ingame in the beginning of the lure time I don't see yours being better.

Config: This would be as good as ingame endtime and would just need a little maintenance.
RAW: This maybe easier to automate, but would also consume scanner time in events.

krzee:
thats a great idea sn0opy! maybe its possible to throw in a clicked stop after event is detected so then we know the event timing, which could be assumed until a stop is detected to un-lure [de-lure?] faster

JimBim89:
Even now there are really no further fields needed, if your assumed end time has passed and an active lure is still reported (with the same modified date) you have a special duration. Without the addition of the methods I mentioned above maps could just show start / end time while the end time in not reached and start / "Extended Lure Time" or whatever if the end time is surpassed and put on a auto-hide after x minutes or wait until its scanned again without the lure active. The last updated field is never under 5 min from previous scan for me, and if you can't detect stops regularly maybe this feature isn't that suitable for your setup if people still need to go to it and you have nearly no lure time left.

bbdoc:
@sn0opy Issue with your pseudo code is that it will take 30 minutes after an event start to detect the new lure time. So during the first 30 minutes, you are setting wrong end times on all stops and sending wrong information to your webhook for all stops. from my experience, most stops are being lured at the very beginning of events, so globally you'll set wrong information for 99% of them. When even ends, you'll also continue to send 3 hours endtime to new lures although they are back to 30 minutes.

I've never seen Niantic change lure timers in the middle of an hour, so I would suggest to simply clic the first lure pokestop you see at the beginning of every hour, and use the lure time received for all other lure until the next hour. This way you get automatic and accurate timers with minimal clics.

JabLuszko:
> I've never seen Niantic change lure timers in the middle of an hour, so I would suggest to simply clic the first lure pokestop you see at the beginning of every hour, and use the lure time received for all other lure until the next hour. This way you get automatic and accurate timers with minimal clics.

No idea how this works on Niantic end, just throwing here for future research if anyone decides too. We have +30 or +45 timezones in general.

Time zone in India (GMT+5:30)
Time zone in Nepal (GMT+5:45)

krzee:
> Time zone in India (GMT+5:30)
> Time zone in Nepal (GMT+5:45)

damn it, Nepal!

bbdoc:
but at what time are events usually starting (in your timezone) ? If they are starting at xxh00, it doesn't really matter… If they are starting at xxh30 or xxh45 then indeed, would mean there is a need for a config item (scan first stop by default at xxh00, but possibility to adapt it).

cecpk:
Nepal ... Pokémon .... aaaha ...

PR #226, opened 2019-05-31
Avoid 100% CPU load during quest scanning.

eckerse:


cecpk:
rly? Nice :D

ColdShadow80:
Using for 2 days now. Love it.
CPU usage peaks dropped to 20%

PR #231, opened 2019-06-09
Merge: Dev -> Master

cecpk:
@everyone we've got exciting news. Dev has been merged into master with a ton of new and improved things! Unfortunately, we also have some bad news: we don't support windows natively anymore. You now have to use either a Linux VM or WSL. More on that on our wiki: https://mad-docs.readthedocs.io/en/latest/installation/1.getting-started.html#windows

For everyone else. You can now `git pull` on the master branch and update/install new requirements via `pip install --upgrade -r requirements.txt`. Also make sure, you run AT LEAST tesseract version 3.05 on your system. Check your version via `tesseract --version`

Changelog:

* General: much faster startup
* General: detect and click on Google account logins
* General: add leveling option for quest mode
* General: new MITM status page (Endpoint /status/ on mitm Port)
* General: Enable/Disable worker stats (for low performance systems like RPi)
* General: add option to switch between a quick and an optimized route calculation mode
* General: proper graceful shutdown when stopping MAD via CTRL+C
* Internal: better button detection
* Internal: Heavy refactoring to run several parts of MAD in seperated processes for better scalability
* Internal: fix 100% CPU when scanning for quests
* Madmin: add button to trigger mappings reload instead of auto-loading it
* Madmin: public quest page
* Madmin: add confirm dialog for delete settings
* Map: view mon
* Map: view priority queue
* Map: view "cell updates". Basically an option to see L15 s2cells that are being retrieved by devices
* Webhook: add forms to quest mon
* Webhook: add lure modules


cecpk:
Tick Tack 🕰

PR #235, opened 2019-06-15
New feature: Scan the "nearby" section to build up a special kind of queue

sn0opy:
*HEAVILY BROKEN RIGHT NOW! 2019-07-02*

This PR will add a new kind of scan mode which actively scans the geofence only for "nearby" mon. Those mon have absolutely no value except you know that they exist. We have no despawn time, no actual position, no spawnpoint ID. Basically nothing. We only get the stop id, the mon id and the encounter id.

This mode can be used to build up a queue to scan for rare mon around stops. Not more. At least that's the idea. Further idea would be to get the spawn point's despawn time around a stop to generate a proper queue.

**Features so far:**
- the new table will be propagated by every scan mode
- New scan mode that quickly generates a route that should see all nearby mon
- Display nearby mon on the MADmin map scattered around their stop

**Bugs / not working / untested:**
- the actual scan mode is untested. It does properly generate a route tho and it does walk around but it might break at some point
- no Monocle support
- map is super laggy
- no automated creation of the new required DB table 
- once propagated, data won't be deleted afterwards
- the map also displays already expired data

**Todo:**
- fix things from list above
- find a proper way to remove old data from the table. (remove all entries from a cell just before re-adding new ones would be one option)
- more like a general todo: add row based route calculation as an option to use

The following table is required for this PR to work properly 
```sql
CREATE TABLE `trs_nearbymon` (
  `cell_id` bigint(20) unsigned NOT NULL,
  `encounter_id` bigint(20) NOT NULL,
  `stop_id` varchar(50) COLLATE utf8mb4_unicode_ci NULL,
  `mon_id` int(11) NOT NULL,
  `detection_timestamp` int(11) NOT NULL,
  PRIMARY KEY (`cell_id`,`encounter_id`),
  KEY `stop_id` (`stop_id`),
  KEY `cell_id` (`cell_id`),
  KEY `encounter_id` (`encounter_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```


m3g4mInD:
@sn0opy can  you add the table & columns need on trs_nearbymon? tks i would try now :)

PR #238, opened 2019-06-17
Added gender support for raid bosses

JimBim89:
Added gender support for webhook type "raid" which is supported by PA for example.
Also modified rm wrapper to pick up information to fill the needed gender data.

This additional gender information for raids is useful for Burmy, Snorunt etc.

### DIFF
--- a/db/monocleWrapper.py
+++ b/db/monocleWrapper.py
@@ -38,6 +38,11 @@ class MonocleWrapper(DbWrapperBase):
                 "ctype": "tinyint(1) NULL"
             },
             {
+                "table": "raids",
+                "column": "gender",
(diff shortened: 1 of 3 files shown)


sn0opy:
Thank you so far. Could you also add this for Monocle please?

JimBim89:
Sure, and proper column creation would be nice I guess.

sn0opy:
I was expecting the gender to be in the raid table already. Looks like it isn't. So yes, please add a proper way to update the tables as well. Have a look at the following section which should add new collumns quite conveniently https://github.com/Map-A-Droid/MAD/blob/master/db/rmWrapper.py#L24-L36

PR #254, opened 2019-06-29
Set lure to None by default.

daangroot:
There is no need to set the lure expiration to a specific date since lure_expiration is allowed to be set to null in the db.


cecpk:
No errors with that?

daangroot:
Didn't test it, but `lure_expiration` is only used when sending pokestop data to a webhook, so it should be fine.

PR #255, opened 2019-07-01
make mappings.json file/path configurable

crhbetz:
Adds optional `-mf` / `--mappings` argument to make it configurable like config.ini (`-cf` / `--config`). Should make it possible to run multiple completely independent instances of MAD out of a single directory.

Currently running this without issue with `-mf` set to a custom file.
Didn't test running multiple instances with different mappings file yet.


crhbetz:
fyi - I've been running two separate instances out of the same MAD folder with this for two days now - separate config and mappings files - and haven't seen any issues.

PR #260, opened 2019-07-10
Custom Google Apps domain support in GGL

JabLuszko:
Should work, someone double check please :D


cecpk:
maybe as an array if more diff domains are in use? :)

JabLuszko:
@cecpk there you go, comma sep list

cecpk:
we wait for #267
already steal code from this 🥇

cecpk:
see https://github.com/Map-A-Droid/MAD/commit/7bead975bee2ea11dc16c5712e8edb18eeb9e5f4

PR #261, opened 2019-07-10
Updated link to cleandb.sql

clburlison:


sn0opy:
That paragraph could be removed anyway, since MAD has both SQL schemas shipped within the code including a script to execute them

sn0opy:
Going to close this as I'm going to remove that part from the Readme anway.

PR #266, opened 2019-07-13
Update Dockerfile

Breee:
This is the Dockerfile we currently use to Build a docker-image for MAD.

As you see the dockerfile has two parts:
(1) In the first stage we have to compile tesseract, because the 3.7-slim python image has only access to tesseract 3.04 via apt. 
(2) In the second stage we copy the compiled binary and libaries  to our final image and then install all required packages and requirements to run MAD, finally we throw away everything we do not need to reduce image size even further.

We try to minimize the amount of layers, because the python-requirements and Open-CV blow up the image size drastically already. 
We end up with a final size of  916MB.

If you see (a) an opportunity to reduce the image size even further or (b) even come up with a solution to run it with an alpine base or (c) spot any bugs -  feel free to inform me :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,32 +1,54 @@
-# Basic docker image 
-# Usage:
-#   docker build -t pogomad .             # Build the docker Image
-#   docker run -d pogomad start.py        # Launch Server
+###########################
+# compile tesseract.
+###########################
+FROM python:3.7-slim AS tesseract_builder
+RUN apt-get update && apt-get install -y automake ca-certificates g++ git libtool libleptonica-dev make pkg-config
+RUN git clone https://github.com/tesseract-ocr/tesseract.git && cd tesseract && ./autogen.sh && ./configure && make &&  make install && ldconfig
 
-FROM python:3.7.1-slim
 
-# Default ports for PogoDroid, RGC and MAdmin
-EXPOSE 8080 8000 5000
+############################
+# MAD
+############################
+FROM python:3.7-slim
+# Copy stuff from buildstage
+COPY --from=tesseract_builder /usr/local/bin/tesseract /usr/local/bin/tesseract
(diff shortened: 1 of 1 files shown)


muckelba:
I've tried to fiddle around with apline + MAD but I wasn't able to pass every python package installation error 🤷🏼‍♂️ and building took ages anyway

Breee:
> I've tried to fiddle around with apline + MAD but I wasn't able to pass every python package installation error 🤷🏼‍♂️ and building took ages anyway

The main problem was open-cv when we tried it + some random missing libs which we would have to take from another apks on top of the installation errors you mentioned. 

Also, one has to compile opencv  and copy the binaries into the alpine container, similar as we do with the tesseract binaries - which probably takes ages. This is the only way I can think of how it works. 
Something similar as this guy did maybe:  https://github.com/julianbei/alpine-opencv-microimage

However, I do not think it is worth the effort  for a few MB image size + a bit more safety. 

The python requirements add like 400 mb to the image. Other libaries add the remaining mb. 
Maybe there are some more libaries / packages we can throw away.

tinbromide:
You can install newer versions of tesseract from the stretch backport. Try putting these lines in your dockerfile:

`RUN printf "deb http://httpredir.debian.org/debian stretch-backports main non-free\ndeb-src http://httpredir.debian.org/debian stretch-backports main non-free" > /etc/apt/sources.list.d/backports.list
RUN apt-get update && apt-get -y --allow-unauthenticated -t stretch-backports install tesseract-ocr libtesseract-dev`

cecpk:
🕚 ?

Breee:
Worked in @tinbromide's proposal. Everything works :+1:

georgeherby:
I thought I would test this, it won't build for me...

```
The following packages have unmet dependencies:
 tesseract-ocr : Depends: libicu57 (>= 57.1-1~) but it is not installable
```

Full log:
https://pastebin.com/qgAsLQLi

More than happy to be told I am an idiot and not running the file correctly 😂

Breee:
@georgeherby's problem is MAC-related. On our Servers/Notebooks it builds fine. 

We pushed the image to https://hub.docker.com/r/breedocker/mad and he could pull + run it on his MAC. So ultimately this should not be a problem.

Also, I dunno what one has to do to build this image on MAC.

PR #275, opened 2019-07-19
Potentially webhook fix

pki0:
Users reporting in Discord that they have missed some 100% IV Mons in their webhook clients. They were _probably_ not sent by MAD.
While reading the code I noticed that the `last_check` in webhookworker.py could cause a small race condition: If the query or the for-loop is slow we could miss Mons.

**Example:**
12:00:00:00: Big webhook preparing to send. Loop is slow because big amount of data.
12:00:01:00: Webhook finally sent. => `last_check` = 12:00:01
But the async scanner found a Mon:
12:00:00:50: Mon inserted into database with `last_modified` = 12:00:00
12:00:10:00: Next webhook preparing to send with Mons between 12:00:01 and 12:00:10.
=> Mon from 12:00:00:50 is missing.

~~So I added a 5 seconds interval to the query. Maybe this is too much and we can change it to 1-2 seconds.
Other method could be to modify the `tsdt`. But for me this is a more readable solution.~~

I also removed two unnecessary IF conditions to speed up the webhook-preparing loop.

~~Maybe this could also be a problem with raids and quests.~~

Feel free to merge, close or discuss.

Edit: Now it should work in a more simple way. The function can be slow as it wants, we'll not missing any data.


temp666:
interval is a bad idea. what if when it use more time?
I prefer to move the last check time before it's sent the webhook:
https://github.com/Map-A-Droid/MAD/blob/68d9f779e8f3fc7bc495a881d980f544ab407ee2/webhook/webhookworker.py#L595 to https://github.com/Map-A-Droid/MAD/blob/68d9f779e8f3fc7bc495a881d980f544ab407ee2/webhook/webhookworker.py#L591

with the solution the webhook can use as long as it wants

pki0:
@temp666 
Will it work?
`last_check` is used inside the prepare function: https://github.com/Map-A-Droid/MAD/blob/68d9f779e8f3fc7bc495a881d980f544ab407ee2/webhook/webhookworker.py#L540

Wouldn't this cause an empty payload?

temp666:
yes it's will work. `__last_check` is used in https://github.com/Map-A-Droid/MAD/blob/68d9f779e8f3fc7bc495a881d980f544ab407ee2/webhook/webhookworker.py#L590
and not in https://github.com/Map-A-Droid/MAD/blob/68d9f779e8f3fc7bc495a881d980f544ab407ee2/webhook/webhookworker.py#L593

pki0:
Thank @temp666 for the suggestions. Now it should be correct.

PR #282, opened 2019-07-23
Fix for malformed pokemon after weather/event changes

JimBim89:
This fixes most of the malformed pokemon after event and/or weather changes. For example the falsely boosted community event mons after the event hours have ended.


pki0:
I think we need here more changes if a Pokemon changes but not getting encountered.
In this case we could get something like a 15/15/15 Alolan Magikarp.

So we need to reset all values before the encounter.

JimBim89:
You are right about that but only in monocle, I don't use it so it slipped through.
The form is already updated in rm so I didn't check for that in the other wrapper.

More feedback?

JimBim89:
Monocle should also be up to date now.

PR #284, opened 2019-07-23
Extend trash matching range

rfuchs:
On an xperia tablet, the height cutoff for looking for trashcans is too short and it fails to find the 3rd trashcan on the screen. Extend the range a little bit so it can find the 3rd trashcan.


cecpk:
ok for me - close this. added to dev with last push

ColdShadow80:
NOT OK - Now Moto G4+ does not delete items
[ WARNING] Find no item to delete - scrolling (2 times)

PR #287, opened 2019-07-24
Handle protocol type 156 to update gym details

rfuchs:
This makes it possible to open a gym and automatically populate the
details (name, url, and description) of the gym in the DB.

### DIFF
--- a/db/dbWrapperBase.py
+++ b/db/dbWrapperBase.py
@@ -340,6 +340,13 @@ class DbWrapperBase(ABC):
         pass
 
     @abstractmethod
+    def submit_gym_proto(self, origin, map_proto):
+        """
+        Update gyms from a map_proto dict
+        """
+        pass
+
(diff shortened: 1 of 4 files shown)


cecpk:
Round 1 - fight

muckelba:
Considering [this](https://mad-docs.readthedocs.io/en/latest/common-issues/faq/#is-it-planned-to-see-gym-defenders-trainers-in-gyms) statement, i think name, url and description should be fine. But your PR doesn't has any automatic mechanic to click on the gyms, so the benefits right now is not *that* huge. Still a good QoL update :)

rfuchs:
For a new installation, it was easier for me to do it this way than to manually fill in the details :)

Automatic clicking on gyms can be added as a second step.

muckelba:
Btw, we have an import script for that :D

PR #289, opened 2019-07-24
add shebang to databasesetup.py

JabLuszko:
shebang please fix

### DIFF
--- a/scripts/databasesetup.py
+++ b/scripts/databasesetup.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 import sys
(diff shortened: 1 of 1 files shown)


sn0opy:
This wouldn't change anything, if the script is not exectuable

JabLuszko:
Someone on discord made it executable and was greeted with bash trying to run python script. This way maybe it will work for someone. No cons here imho.

sn0opy:
Well. you have to do both. Make it executable AND add the shebang. This only adds the sheband, so one still has to make it exectuable manually.

PR #292, opened 2019-07-26
 Allow RM name column from gymdetails table to be NULL.

daangroot:
Now we no longer have to insert 'unkown' for every gym.
This time I've fully tested it ;) Db migration is present aswell.


temp666:
what about scripts like this:
https://github.com/Map-A-Droid/MAD/blob/dev/scripts/update_stopsgyms.sh

daangroot:
@temp666 Has been updated: https://github.com/Map-A-Droid/MAD/pull/292/files#diff-7176978b96f15b2769cc2d4f43e69d17R84

sn0opy:
How does RM handle these things nowadays? Is it still requested, to have this set to NULL instead? Would you be so kind to re-do the PR directly against master?

PR #296, opened 2019-07-27
Respell "Statisics" in statistics.py

granis:
Replaced 2 occurrences of "Statisics" with "Statistics".
No OCD at all.


granis:
Completely forgot to submit to dev, sorry, ignore this :D

PR #298, opened 2019-07-29
Added grunt type to db and pokestop webhook

JimBim89:
Added grunt type support to database and webhook type "pokestop".
Also modified rm wrapper to pick up information to fill the needed data.

This additional type information for invasions is useful to hunt for specific mon types.

This uses the same naming scheme as rdm / pmfs for easier compatibility.

### DIFF
--- a/db/rmWrapper.py
+++ b/db/rmWrapper.py
@@ -56,7 +56,7 @@ class RmWrapper(DbWrapperBase):
             {
                 "table": "pokestop",
                 "column": "incident_grunt_type",
-                "ctype": "tinyint(1) NULL"
+                "ctype": "smallint(1) NULL"
             }
         ]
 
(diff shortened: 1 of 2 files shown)


sn0opy:
please prefix `grunt_type` with "incident_" to make it clear they belong together

db/rmWrapper.py
```diff
@@ -52,6 +52,11 @@ def __ensure_columns_exist(self):
                 "table": "pokestop",
                 "column": "incident_expiration",
                 "ctype": "datetime NULL"
+            },
+            {
+                "table": "pokestop",
+                "column": "grunt_type",
+                "ctype": "tinyint(1) NULL"
```
fosJoddie: is tinyint enough here? if I remember right it has a value of max 127 - maybe it's overkill with smallint here, but who knows how many other types of character displays they might add in the future.

JimBim89:
Pushed the requested changes!

PR #301, opened 2019-07-29
sort mappings entries alphabetically in madmin view

JabLuszko:
Areas, devices, walkers, auth and global mon list


cecpk:
rly? :D

sn0opy:
I love it. Tried the same with all the stuff that is being sent to the madmin map causing the CPU to freak out. Looks like this ia proper solution.

georgeherby:
LGTM

PR #305, opened 2019-07-31
update Dockerfile

Breee:
Fixed an issue where we override the apt sources lists, which breaks apt.
Technically this was no problem for MAD, rather a QOL improvement when working inside the Container. (in case you want to install something via apt)


sn0opy:
Same changes as in #316.

PR #311, opened 2019-08-01
Rename Phonecontrol to Devicecontrol

muckelba:
I don't know if that's a useful PR but i thought it's more correct to use the term `device` instead of `phone` considering the increased popularity of ATVs

But maybe its just my inner monk and nobody cares about it :D if that's the case: just close the PR ^^

### DIFF
--- a/configs/config.ini.example
+++ b/configs/config.ini.example
@@ -82,7 +82,7 @@
 
 # ADB
 ######################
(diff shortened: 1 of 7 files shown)


sn0opy:
Going to put this on hold until madmin rework has been merged

sn0opy:
Probably a good idea to re-do this PR since there have been so many changes since then. Feel free to PR this directly against master

PR #315, opened 2019-08-05
Change route calculation to consider S2 cell level for viewing radius…

eckerse:
For now only active for raid_mitm route calculation. Other modes still disabled, but should work also

uses s2sqhere regioncoverer to get all cells touched by radius and checks for each position if it lies inside those cells.

Should also work with other modes if correct radius and s2level are chosen.


sn0opy:
So, is this something that actually works? Did others test this already? I was reading a few things about it on Discord, but haven't had the time to test this myself yet.

eckerse:
Didn't get feedback yet. For me it works good (before change route had a size of approx 95 coords, after using this calculation method reduced to approx. 82, result will depend on gym density of course). According to cellupdates in madmin map I don't miss any cells.

Important change is in ClusteringHelper.py and s2Helper, most other changes are only to make S2 based calculation be optional.
It will use S2 calculation for raid mode only, all other methods will use old method by default. It should work for mon mode also, but I didn't check the cell size of mon detection (S2L17?)

hzpz:
I can confirm, that with this change, MAD calculates routes with fewer coords. Judging from the "last scanned" timestamp it seems that MAD still sees all gyms and I know from personal experience that it is possible to see all gyms from fewer coords than MAD used before. So yes, it seems to actually work 👍

eckerse:
Added a new PR #475 based on current master as this one needed some manual merging.

PR #318, opened 2019-08-09
Ability to send non-iv'd mons (that are on iv list) via webhook.

JabLuszko:
By default MAD do not send non-iv'd pokemon via webhook if they are on ANY IV list.
It's still marked as #TODO and I don't want to mess with @sn0opy vision how to do it - this way we can safely enable sending non-IV'd mons to webhook, I guess.

Default behavior is not changed - user must uncomment a line in config.ini to have it working.

**WARNING**: Your webhook receiver must support getting same encounter twice - once with IVs, once without IVs. For example PokeAlarm does this only in dev branch.


LucianoNeo:
pleaaaaaaaaaaaaaaaaaaaaaaseeeeeeeeeeeeeeeeeeeeeeeeeee

Issue #329, opened 2019-08-15
MADmin geofence editor coordinates go out of bounds.


sebastienvercammen:
When using MADmin's map to draw a geofence, if you've scrolled past the bounds of the original world, the saved coordinates will be beyond the bounds of the real world, causing the route/position generation to go beyond the geofence and map the world.

**Environment:**
* `dev` branch, latest commit (https://github.com/Map-A-Droid/MAD/commit/0d97dc25f9680b75cf1821dc96ff17511319e37a)
* py3.7
* Raspbian Buster

**To fix:**
Modulo. Keep coordinates within bounds.
_(Keep in mind the difference between remainder [= JS' modulo] and [least positive residue](https://en.wikipedia.org/wiki/Modulo_operation#Remainder_calculation_for_the_modulo_operation) [= python's modulo].)_

**Steps to reproduce:**
1. Open MADmin's map.
2. To make things easier/quicker, zoom all the way out till you see the world multiple times. _(Step can be omitted.)_
3. Move the map beyond the bounds of the original world.
4. Draw a geofence in any world that is not the very first one.
5. Save the geofence.

The saved geofence will have coordinates within the bounds of the world you drew it in, beyond the bounds of the first world.

The problem: When generating the route (positions), it will try to generate routes to map the world several times, going beyond the geofence. No error message or crash, just stuck generating positions.

fosJoddie:
I got the fix for this, will PR as soon as I get the time

PR #335, opened 2019-08-18
Multiworld support when saving geofences from madmin

fosJoddie:
Leaflet handles very zoomed out views by simply adding multiple worlds. This causes coordinates from it to not always be inside [90, 360] - this fix makes sure we don't duck up saving of geofences.


fosJoddie:
Fixes issue #329

sn0opy:
Not entirely sure, if we should just ignore this for now. We plan to move everything into the DB anyway + people normally use MAD  with basic authentication anyway

fosJoddie:
just an FYI this PR does not introduce this error.

Furthermore this change would have to be in the logic that saves to DB as well to prevent the same error. + it's a fix that is ready right now - vs the "save in db" is a ways off

PR #339, opened 2019-08-19
[WIP] New shiny stats with heatmap.

JabLuszko:
Link in dropdown - Shiny stats.
Not "per worker" ratio anymore, that was slow (and imho useless, sorry cec ;) ).

Totally Beta. Could be totally broken or even more. Try, report back.


JabLuszko:
Don't use it for now :) This version is mostly for Duke.

Currently there is new column in pokemon table (in RM, mono is dead 😄) and it tries to update pokemon table with current data from raw table. It will take time on start and changes your table schema.

There are still things I don't like and have no idea how to approach this atm - like last_modified is getting updated every time MAD sees mon (even from overwold), so it's not really as good/same as timestamp_scan from raw. Using disappear_time from pokemon will have nothing to do with time of scan either. Using disappear_time minus spawn length will be wrong for CD mons. Extra column is probably totally not worth it, we should just do not delete entries from raw table if they contains info about shiny.

cecpk:
i want the headmap!

JabLuszko:
I've adjusted cleanup statistic to NOT delete entries with is_shiny=1. I don't think there is gonna be a lot of entries (especially with how many entries this thing generates), but dropping all shiny stats after one week is kinda meh. I hope that is okey?

Back to using trs_stats_detect_raw for shiny stats.

There are two /d/ columns for debug purposes - I have strange data (as in two shiny from same worker at the same time 600 meters afar) in my trs_stats_detect_raw and this will probably need looking into it.

Not deleting old links for now.

@cecpk poor man's heatmap ready 🐤

sn0opy:
@JabLuszko is this ready to merge? We're preparing a master merge and I think having this on master would be another nice addition.

JabLuszko:
Works for me, need more feedback... so dev ready I guess (?)

sn0opy:
Alright, two things:

1. could you remove your disclaimer/info at the bottom of the page?
2. would it be a good idea to use the average or even better the median for the shiny hour stats? Last CD broken the stats for me for example:

image

JabLuszko:
1. Is gone.
2. Can't really imagine how to use mediana here - but I am totally open to changing anything here - it's gonna sky rocket over time with more and more scans/shiny. It kinda worked with 7 days reset, but now data should be saved forever.

PR #348, opened 2019-08-22
Add new quest types to the quest parser and EN languages

fosJoddie:
Adds new quest types like grunt battles and take snapshots that are going to be used in Water Festival

PS - missing DE and FR translations

NEEDS TESTING


cecpk:
DE and FR later

PR #349, opened 2019-08-23
Fix oversight in spinnable check

fosJoddie:
If spinnable check fails in first try, we add stop to ignore list ONLY if we didn't get a new GMO.

This causes clients that are bugged out (player char is gone from overworld view + no map features visible) to simply start ignoring any and all pokestops instead of timing out and restarting pogo after N amount of retries.

This *should* reduce amount of stops we skip because of a bugged out worker by only considering stops to be ignored IF we got a GMO AND the stop wasn't in that GMO.


cecpk:
added to dev - i close this now

PR #355, opened 2019-08-24
Add french translation of new quest

na-ji:
I updated the french translation from the .pot file


cecpk:
thx - merged in my last push

PR #356, opened 2019-08-24
make the max injection count variable through a setting in config.ini

bk227865:
In stead of a hardcoded number of injection tries (20) create a variable in config ini that can adjust the number of injection tries before rebooting.


cecpk:
#359 is better - i close this

PR #360, opened 2019-08-26
Restart POGO from madmin phonecontrol

JabLuszko:
Same button like last time, just more options after click.
Image

Copy-paste of ajax calls, probably should be moved to dedicated method soon.

Not sure if ADB part works - can't test. It should, in theory.


darkelement1987:
maybe change the 'Quit Pogo' text to something like Quit/Restart Pogo?

PR #361, opened 2019-08-27
Update vars_parser.json

kamieniarz:
Previous description was a bit shitty xd


darkelement1987:
thank you, i was to lazy to pr

PR #362, opened 2019-08-29
Fix madmin geofence editor lat,lat

JabLuszko:
Save lat, lon not lat, lat.

On top of that please also merge #346 
Confirmed working with those changes :)


sebastienvercammen:
Related #344 

Don't merge #346. Arbitrarily removing the last item is not a real solution. Instead, just skip empty `latlon_split` items.

PR #366, opened 2019-08-31
check for pogo freeze

cecpk:
detect same screens while screendetection


darkelement1987:
Is this ready to merge 😁?

PR #367, opened 2019-09-02
Fix saving geofence coords for negative values

JabLuszko:
There are bounds there in map definition now, there is no need for modulo 360/90.
Btw. it was broken before too.
```
>>> 80%360
80
>>> -80%360
280
```


fosJoddie:
I swear this is the last time I phone in a "quick fix" for something and then nobody bothering to actually test it... 🤣 thanks for your fixes man

PR #371, opened 2019-09-03
Add trs_s2cells table to rocketmap.sql

clburlison:
I created a new instance from scratch for testing and kept getting errors since the trs_s2cells table didn't exist.

```bash
09-03 17:01:57.05] [   MITMReceiver-0] [      dbWrapperBase:174 ] [   ERROR] Failed executing query: 1146 (42S02): Table 'mad.trs_s2cells' doesn't exist
```

### DIFF
--- a/scripts/SQL/rocketmap.sql
+++ b/scripts/SQL/rocketmap.sql
@@ -303,6 +303,14 @@ CREATE TABLE `trs_quest` (
   KEY `quest_type` (`quest_type`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
 
+CREATE TABLE `trs_s2cells` (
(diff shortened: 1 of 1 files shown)


muckelba:
Actually, a couple tables are missing, but that is not a problem usually, since MAD will create them at start.

clburlison:
Right but this specific table wasn't being created by MAD

JabLuszko:
https://github.com/Map-A-Droid/MAD/blob/master/utils/version.py#L259

muckelba:
Sounds more like a MAD issue in general. I'm not sure how MAD currently behaves at the first startup, but it will add that table at the second run. (And/or after deleting version.json and restarting MAD)

Cc @sn0opy

clburlison:
I see my mistake now. I copied my prod setup and pasted it to a new directory. That's why the version.json file existed and why the database setup didn't complete correctly.

I know for next time. Thanks!

PR #373, opened 2019-09-05
Show sleep time in status page

fosJoddie:
Allows (quest) workers to save the current sleep time in status table so it can be seen in madmin worker status page like this: 

image

Why? Because then you can see if it's actually stuck or just on a long sleep time between areas.

Tested a bit, but more tested needed 🙏 

Also fixes some missing columns in the code for new dbs `trs_status` table

### DIFF
--- a/db/dbWrapperBase.py
+++ b/db/dbWrapperBase.py
@@ -839,7 +839,7 @@ class DbWrapperBase(ABC):
                  ' globalrestartcount INT(11) NULL DEFAULT 0, '
                  ' lastPogoReboot VARCHAR(50) NULL DEFAULT NULL , '
                  ' globalrebootcount INT(11) NULL DEFAULT 0, '
-                 ' currentSleepTime INT(11) NOT NULL DEFAULT 0'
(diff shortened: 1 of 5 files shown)


cecpk:
❤️

db/dbWrapperBase.py
```diff
@@ -836,6 +836,10 @@ def create_status_database_if_not_exists(self):
                  ' init TEXT NOT NULL, '
                  ' rebootingOption TEXT NOT NULL, '
                  ' restartCounter TEXT NOT NULL, '
+                 ' globalrestartcount INT(11) NULL DEFAULT 0, '
+                 ' lastPogoReboot VARCHAR(50) NULL DEFAULT NULL , '
+                 ' globalrebootcount INT(11) NULL DEFAULT 0, '
+                 ' currentSleepTime INT(11) NOT NULL DEFAULT 0'
```
cecpk: missing "," ?

fosJoddie: Fixed

PR #374, opened 2019-09-06
Raid Pass name correction.

boby360:
Previously this didn't properly reflect the wording used ingame for EN.

This now uses the proper wording used ingame for EN.

THIS IS UNTESTED.


boby360:
This doesn't seem to fix anything despite the current code appearing to be incorrect.

PR #380, opened 2019-09-09
Add option to disable coloring in stdout/err

chr2002qw:
Perhaps useful for those who running MAD by systemd. Color codes make the syslog difficult to read.

Added option -nc (--no_color) to cmdline to disable coloring in stdout/stderr.
MAD can be run by systemd and split logs using syslog into a seperate logfile. So every output logged to stdout/stderr including tracebacks is written to this file.

My first PR. Possibly dirty code ahed :-)

### DIFF
--- a/utils/logging.py
+++ b/utils/logging.py
@@ -11,14 +11,44 @@ def initLogging(args):
     log_trace = log_level <= 10
     log_file_trace = log_file_level <= 10
 
-    logconfig = {
-        "levels": [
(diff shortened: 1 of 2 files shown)


sn0opy:
Properly implemented in https://github.com/Map-A-Droid/MAD/commit/df3f7cea837bc9d876466e9f3292d8f2ed85459f. Thanks anyway!

New Option: `--log_colors`. Default: true

PR #381, opened 2019-09-10
Missing param for generating route

JabLuszko:
I guess it should be like that :)


fosJoddie:
Seems to do the trick

PR #392, opened 2019-09-18
Magiskhide : Let MAD remove / add Pogo from Magiskhide

chr2002qw:
Works on screentype.FAILURE and screentype.RETRY.

Stop PoGo - remove from Magiskhide - add to Magiskhide - start Pogo.

Could help preventing the "not compatible" screen.

Code not as clean. Please review carafully :-)


sn0opy:
I actually never had to do this to fix my devices. In what way does this help? The "incompatible" screen may appear every now and then, but MAD will just restart my device and the problem is gone.

crhbetz:
It's faster than a reboot and also very reliable. Used to do this externally and it worked really well.

acocalypso:
This breaks madmin in master

root@v22019051561290579:~/MAD# python3 start.py -vvv
  File "start.py", line 30, in <module>
    from websocket.WebsocketServer import WebsocketServer
  File "/root/MAD/websocket/WebsocketServer.py", line 23, in <module>
    from worker.WorkerMITM import WorkerMITM
  File "/root/MAD/worker/WorkerMITM.py", line 13, in <module>
    from worker.MITMBase import MITMBase, LatestReceivedType
  File "/root/MAD/worker/MITMBase.py", line 13, in <module>
    from worker.WorkerBase import WorkerBase
  File "/root/MAD/worker/WorkerBase.py", line 641
    self._stop_pogo()
    ^
IndentationError: unexpected indent
root@v22019051561290579:~/MAD#

darkelement1987:
Fixed with #401

PR #393, opened 2019-09-20
TEST PR DO NOT CONSUME

fosJoddie:
make devices actually editable

add shared settings

make areas configurable

make 'reload settings' work

some fixes

another fix

Implemented the API for MADmin.  This update includes changing mappings.json to be a more db-like structure

Updated the front-end to use the API for fetching data.  Monlist and MappingManager have not been updated to use the new structure

Moved monlist vars into the configuration

Contents from vars_parser.json has been moved into their respective API classes

Removed functions used by the old frontend

Converted the configuration to by python versus json

Temporary file used for launching MADmin until mapping manager has been initialized

Removing the run-madmin file.  Will put into gist

Created a data interface that allows for MADmin to make local calls for fetching data.  API has been updated to pull from this interface

Updating monivlist to match the configuration

Fixed an issue where you could not create a new resource

Updated configmode.py to use the data manager

Implemented the monivlist page

Updated MappingManager to use the DataManager for pulling data

Dependency checking at the data manager level has been added.  The UI will now display any dependency issues

Updated the main startup script to include the data manager

MADmin map now displays the correct information

Updated mappings_example to reflect the API changes

Update to the conversion to account for a section may not exist

A few MADmin fixes

Advanced configuration mode for allowing the user to modify the settings via the UI

Correct implementation for Areas handling mode

Updates to the area configuration to display the same information as the old UI

The empty value from the settings is now being respected

Enforcing walkervalue when creating a walkerarea

Updated the RouteManagers to use the DataManager

Fixed how the conversion works for the initial upgrade around ivlists.

Change the order in formatting an incoming value

Pokemon IV List correctly saves and can have elements removed from the list

FIX: version.py for mappings.json converter

FIX: init of areas

Pokemon list picker has been implemented (#390)

* Pokemon list picker has been implemented

* Fixed an issue where the RouteManager was not correctly initialized

* Updated default redirects

* apiHandler formatting general exception handled

* Updated default landing pages


hartlomiej:
nom nom nom 😋😋😋

PR #396, opened 2019-09-25
Check max. walkers per area separated by walker

cecpk:


sn0opy:
Going to close this for now to revisit this at a later state. Still covered by #428

PR #399, opened 2019-09-26
Toggle magisk on/off after two loginfail screens

chr2002qw:
PR´d this now against master.

Since i'am using this PR, i never looped in loginfail screens.
Added reboot for besser sucess.

Includes Grennith´s patch :-)


Grennith:
Use the patch file I sent please, added methods to communicator won't be executed like that, workers will crash

PR #404, opened 2019-09-29
Display IV0 on madmin

korentzari:
IV0 is not iv not scan

### DIFF
--- a/madmin/static/js/madmin.js
+++ b/madmin/static/js/madmin.js
@@ -1159,7 +1159,7 @@ new Vue({
       }
 
       var ivtext = "";
-      if (iv !== 0) {
+      if (mon["cp"]  > 0) {
         ivtext = `
             <div class="iv">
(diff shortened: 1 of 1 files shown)


darkelement1987:
Can you give a better description what this PR Will change?

korentzari:
MADAMIN not display information when pokemon is IV0 ( atk=0 , def=0 , sta=0 ) 
it is more relevant to test another parameter than attack, defense, stamina which can be zero.

IV0 != NotDetectIV

darkelement1987:
Some people like 0% mons for weird reasons

muddpunk:
I collect 0% Mon. Don't deny me a zero 👀

ColdShadow80:
No need for 0% stats, even colectors sould not be using MADadmin to get notifications of wen they apear, use frontend

hartlomiej:
@darkelement1987 @muddpunk @ColdShadow80 check what it does before commenting
@korentzari good PR

PR #406, opened 2019-10-02
another way to detect ditto

Tabbomat:
Currently, MAD treats all Pokemon as Ditto that have weather boost, but have IV or level that contradicts weather boost.

Pokemon without weather boost that have a level that is only possible with weather boost are also Ditto.

TL;DR: level > 30 and no weather boost means Ditto.

Advice on formatting is welcome as this is my first PR.


ColdShadow80:
On the 2 Dittos the webhook reported aoive level 30, a Ditto was found but Ditto stats did not match webhook.
For example, webhook data vs field data:

REPORTED: level 34 Ditto CP 664 IV 62.22% 6/11/11
FOUND: level 29 Ditto CP 601 IV 42% 10/5/4

REPORTED: level 33 Ditto CP 756 IV 84.44% 14/13/11
FOUND: level 28 Ditto CP 628 IV 80% 11/13/12

darkelement1987:
Were both scan and player LV30?

JabLuszko:
I think we (I personally did, someone else too :D) asked in chat that webhook will report *actual* mon CP - so when you end with 3 gulpins around you despawning within 30 seconds you can check CP to know which one you need to catch and from which you can run. 

I would guess most people hunt Ditto for quest progress, not to have 100 IVs.

ColdShadow80:
> I think we (I personally did, someone else too :D) asked in chat that webhook will report _actual_ mon CP - so when you end with 3 gulpins around you despawning within 30 seconds you can check CP to know which one you need to catch and from which you can run.
> 
> I would guess most people hunt Ditto for quest progress, not to have 100 IVs.
> 
> And this discussion is well outside of scope of this PR btw.

Just provided feedback. In short, PR seems to work. Pokemon data sent to webhook is the stats of the disguise.
Expect a Level reduction (by 5 levels aparently) after catching the Ditto flagged by this PR.

sn0opy:
I'm not that much into the game mechanic here. Is this still something we need? What would it do better over the current way MAD works?

JabLuszko:
> I'm not that much into the game mechanic here. Is this still something we need? What would it do better over the current way MAD works?

This PR will just detect more Dittos (other edge case) than what we have already. it's the same approach used in first place (for low level mons), but adjusted for high level mons. Useful.

Most talk here is generate because people want to know real Ditto stats, not the mon stats that Ditto is disguised by... but without more fields in db we must stick to one mon stats - Ditto or Disguise.

PR #408, opened 2019-10-07
Mention that authentication is for both RGC and Pogodroid

fosJoddie:
Minor text fix for clarification


sn0opy:
Already handled in #411

PR #415, opened 2019-10-13
Add forms and costumes to webhook payload

darkelement1987:
Tested with https://github.com/PokeAlarm/PokeAlarm/pull/694

image


darkelement1987:
Added costume payload as well

darkelement1987:
ok this was unnecessary lol

PR #416, opened 2019-10-14
Wait for pokestop in GMO even for levelmode

fosJoddie:
Levelmode might skip a lot of stops because we do not wait to check for pokestop in GMO before we consider a stop non-existant. Especially bad when game has froze because of long teleport distances


fosJoddie:
this requires more work :o

PR #422, opened 2019-10-17
Real levelup mode that remembers what is already visited

fosJoddie:
Introducing: A levelup mode that you can restart quickly!

How? - Just sit back and let the magic of databases remember where you've been so route coordinates can be skipped quickly in quest level mode.
note: this works for areas in mode pokestop with "level" set to true.

Caveats: Currently only one worker can visit an area before it is forever tainted as visited. A better version is in the works but requires more work and maybe a new mode.
 
FAQ:
Q "one worker finished up a quest area and now the next worker one just says there's nothing to spin"
A run `DELETE FROM trs_visited WHERE origin'thatoldworkerorigin'` and restart MAD
Q how come there are no more questions
A I run out of time

### DIFF
--- a/db/dbWrapperBase.py
+++ b/db/dbWrapperBase.py
@@ -423,10 +423,9 @@ class DbWrapperBase(ABC):
         return spawnret
 
     def submit_pokestop_visited(self, origin, latitude, longitude):
-        logger.info("Flag pokestop as visited")
+        logger.debug("Flag pokestop as visited...")
         query = "INSERT IGNORE INTO trs_visited SELECT pokestop_id,'{}' " \
                 "FROM pokestop WHERE latitude={} AND longitude={}".format(origin, str(latitude), str(longitude))
-        logger.info("Query: {}", query)
         self.execute(query, commit=True)
 
     def submit_spawnpoints_map_proto(self, origin, map_proto):
--- a/db/dbWrapperBase.py
+++ b/db/dbWrapperBase.py
@@ -422,6 +422,13 @@ class DbWrapperBase(ABC):
             spawnret[int(row[0])] = row[1]
         return spawnret
 
+    def submit_pokestop_visited(self, origin, latitude, longitude):
+        logger.info("Flag pokestop as visited")
+        query = "INSERT IGNORE INTO trs_visited SELECT pokestop_id,'{}' " \
+                "FROM pokestop WHERE latitude={} AND longitude={}".format(origin, str(latitude), str(longitude))
+        logger.info("Query: {}", query)
+        self.execute(query, commit=True)
+
     def submit_spawnpoints_map_proto(self, origin, map_proto):
         logger.debug(
             "DbWrapperBase::submit_spawnpoints_map_proto called with data received by {}", str(origin))
--- a/db/rmWrapper.py
+++ b/db/rmWrapper.py
@@ -5,7 +5,7 @@ from multiprocessing.managers import SyncManager
 from typing import List, Optional
 
 from db.dbWrapperBase import DbWrapperBase
-from utils.collections import Location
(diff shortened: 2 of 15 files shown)


korentzari:
Origin is good improvement . 
With rotation is not allways same .Is not possible get username ou more simple index on rotation  ?

fosJoddie:
I'll look into using accountname, index or something in addition to origin for this. It's a lot of rework to get in there. 

There also needs to be a way to clear out level info for an origin before this can be finished. That still has to be figured out. Perhaps clearing out the info on account rotation would be enough.

ColdShadow80:
Maybe clearing data older than X days on account rotation can be a working shortcut.
Not sure if account rotation can occur for other reasons than reaching level 30.

korentzari:
In past, i have used login with database (before implementation in MAD) . 

Is simple and secure storage 

Each time need send login/username  PTC  

Reset login information for origin 
query = 'UPDATE accounts SET origin = NULL WHERE origin = \'%s\'' % origin

Reserve new login with specific level
query = 'update accounts set origin = \'%s\' ,  attempt = attempt + 1 , lastget = now()  where origin is null and status  > 0 and  lvl >= %s and lvl <= %s order by lastget,lvl desc limit 1' % ( origin , min , max)

And after get credential . 
If no login is disponible previous update not reserve account and return error

And for security and force get new account each attempt choice "Date" and " choice Google/PTC" 
clear origin in database.


```
CREATE TABLE `accounts` (
  `id` int(6) NOT NULL AUTO_INCREMENT,
  `login` varchar(32) COLLATE utf8mb4_unicode_ci NOT NULL,
  `password` varchar(32) COLLATE utf8mb4_unicode_ci NOT NULL,
  `playname` varchar(32) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
  `team` enum('mystic','valor','instinct') COLLATE utf8mb4_unicode_ci DEFAULT NULL,
  `status` int(6) DEFAULT 0,
  `origin` varchar(50) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
  `lvl` int(11) DEFAULT 0,
  `exp` int(11) DEFAULT 0,
  `pokestop` int(11) DEFAULT 0,
  `km_walk` float DEFAULT NULL,
  `lastok` datetime DEFAULT current_timestamp(),
  `lastget` datetime DEFAULT current_timestamp(),
  `lastko` datetime DEFAULT NULL,
  `sb` datetime DEFAULT NULL,
  `attempt` int(11) DEFAULT 0,
  PRIMARY KEY (`id`),
  UNIQUE KEY `login` (`login`)
) ENGINE=InnoDB AUTO_INCREMENT=10157 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci 
```

ChaosEmpire:
Why only remember in levelmode?
I think it would be handy also in other modes to know where the device has been last time, like  dynamic switching of devices to other walkers without getting a softban

or restarting and sending the device to last region instead of ocean, wrong town and so on

the level info could be used to distribute devices to correct areas

like having raid scan , mon mitm , iv mitm same time active in one walker (steered by maxdevices per area)
that way a lvl30 acc with its device could directed to the area where iv is scanned, a below 30 acc can moved to the raid scan and so on

i see lots of possibilities to use this data in daily mad jobs, beside level mode

fosJoddie:
status update: 
Suggestions about further improvements are much welcome, but well outside the scope of what I am 
aiming for in this first release. 

TODO:
Refactor my code to only keep the part that is special for level up.

Currently there are two level modes:
* level mode with ignore spun stops
* level mode without ignore spun stops

Either I gotta consider the "ignore spun stops" setting or it has to be remove

Write some sort of function or way to clear "spun pokestop" from DB for a worker

PR #423, opened 2019-10-17
Madmin rework fixes

Expl0dingBanana:
Fix for idle geofence and walkermax not working with '0'.

### DIFF
--- a/madmin/api/modules/ftr_walkerarea.py
+++ b/madmin/api/modules/ftr_walkerarea.py
@@ -43,7 +43,7 @@ class APIWalkerArea(apiHandler.ResourceHandler):
                 "settings": {
                     "type": "text",
                     "require": False,
-                    "empty": None,
+                    "empty": "",
                     "description": "Number of walkers than can be in the area",
                     "lockonedit": False,
                     "expected": int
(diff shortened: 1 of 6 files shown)


madmin/api/modules/ftr_walkerarea.py
```diff
@@ -43,7 +43,7 @@ class APIWalkerArea(apiHandler.ResourceHandler):
                 "settings": {
                     "type": "text",
                     "require": False,
-                    "empty": None,
+                    "empty": "",
                     "description": "Number of walkers than can be in the area",
                     "lockonedit": False,
                     "expected": int
```
clburlison: I'm pretty sure this needs to be changed to a `str` or the int needs to be converted to string type when writing to disk.

Expl0dingBanana:
Looking through the code, this seems to only be referenced in worker/WorkerBase.py.
```
    def check_max_walkers_reached(self):
        walkermax = self._walker.get('walkermax', False)
        if not walkermax:
            return True
        reg_workers = self._mapping_manager.routemanager_get_registered_workers(self._routemanager_name)
        if int(reg_workers) > int(walkermax):
            return False
        return True
```
so that should not cause a problem

clburlison:
👍 you're correct I was impatience on my tests. PR fixes the issue properly.

PR #446, opened 2019-10-20
MADmin behind reverse proxy (non-root)

### LINKED ISSUE #442: madmin doesn't run nicely behind reverse proxy
I currently use an nginx based entrypoint proxy to have SSL support for madmin, RGC and PogoDroid and map these services to 3 virtual locations within my vhost. However, madmin doesn't run nicely as service behind a reverse proxy. 

That's because Flask's (actually werkzeug's) URL handling doesn't support such as setup out of the box. Werkzeug includes the `ProxyFix` middleware to mitigate this problem, but it requires any URL to be created via `Flask.url_for()` which madmin - most of the time - does not.

For the time being, I've worked around this issue with a bunch of freaky rewrite and text substitute directives for an nginx reverse proxy setup, but this is very error prone.

I'd be willing to provide a PR to address the issue. It would generate any URL in madmin with `url_for` and add the `ProxyFix` middleware.

Expl0dingBanana:
This change allows MADmin to work behind a reverse proxy if it is not contained in the root (for example, /madmin) and should resolve #442 .  To use the non-root document, you must specify the walker arg 'madmin_base_path'.

### DIFF
--- a/madmin/api/__init__.py
+++ b/madmin/api/__init__.py
@@ -37,7 +37,7 @@ class APIHandler(object):
             self._modules[tmp.uri_base] = tmp.description
 
     def create_routes(self):
-        self._app.route(self.uri_base, methods=['GET'], endpoint=self.component)(self.process_request)
+        self._app.route(self.uri_base, methods=['GET'], endpoint='api_%s' % (self.component,))(self.process_request)
 
     def process_request(self):
         """ API call to get data """
--- a/madmin/api/apiHandler.py
+++ b/madmin/api/apiHandler.py
@@ -37,10 +37,10 @@ class ResourceHandler(object):
         """ Creates all pertinent routes to for the API resource """
         if self.component:
             route = self.uri_base
-            self._app.route(route, methods=['GET', 'POST'], endpoint=self.component)(self.process_request)
+            self._app.route(route, methods=['GET', 'POST'], endpoint='api_%s' % (self.component,))(self.process_request)
             if self.iterable:
                 route = '%s/<string:identifier>' % (self.uri_base,)
-                self._app.route(route, methods=['DELETE', 'GET', 'PATCH', 'PUT'], endpoint=self.component)(self.process_request)
+                self._app.route(route, methods=['DELETE', 'GET', 'PATCH', 'PUT'], endpoint='api_%s' % (self.component,))(self.process_request)
 
     def format_data(self, data, config, operation):
         save_data = {}
--- a/madmin/madmin.py
+++ b/madmin/madmin.py
@@ -15,6 +15,7 @@ from madmin.routes.map import map
 from madmin.routes.config import config
 from madmin.routes.path import path
 from madmin.api import APIHandler
+from madmin.reverseproxy import ReverseProxied
 
 
 sys.path.append("..")  # Adds higher directory to python modules path.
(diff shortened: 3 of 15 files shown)


Expl0dingBanana:
sample nginx config to run on /madmin on the reverse proxy (no additional configuration required on madmin):
    server {
        listen 80;
        server_name <proxy_url>;
        access_log   /var/log/nginx/mad-dev.access.log  main;

        location ~/madmin(.*)$ {
            proxy_pass http://<mad-host>:5000$1$is_args$args;
            proxy_set_header X-Real-IP  $remote_addr;
            proxy_set_header X-Forwarded-For $remote_addr;
            proxy_set_header Host $host;
            proxy_set_header X-Script-Name /madmin;
            client_max_body_size 100M;
        }
    }

sn0opy:
@chr2002qw @DrMurx please have a look and report any issues you might encounter while testing.

DrMurx:
@Expl0dingBanana @sn0opy Tested everything, works like a charm now. Great job!

muckelba:
Can somebody of you Nginx users either send me a working config or just create a pull request for the Wiki? I think that would be an useful addition. Especially, when a kinda special config is needed.

DrMurx:
This one is supposed to work, replace the `(madmin)` in the `location` directive with whatever route you want (but keep the brackets!).

```
location ~ /(madmin)(.*) {
    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto https;
    proxy_set_header X-Script-Name /$1;
    proxy_http_version 1.1;
    proxy_pass http://<mad-host>:5000$2$is_args$args;
}
```

chr2002qw:
Workes here, too. With config from DrMurx. Thanks 2 all :-)

DrMurx:
@Expl0dingBanana The links from the "Status" page to devices resp. route settings are not covered.

chr2002qw:
This works, Thank you @Expl0dingBanana , @sn0opy :-)

DrMurx:
It works for the actual issue.

Occasionally, however, the fact that the IDs are composed from the path part of the API URI causes compatibility issues - if an object was created prior to this branch, the ID in mappings.json is e.g. `/api/walker/2`, but under some circumstances, the edit form refers to an ID `/subpath/api/walker/2`, so a subsequent PATCH request fails.

This is a related, yet different problem which should be addressed by simply just using numeric values as IDs.

Expl0dingBanana:
DrMurx can you state what steps you took to get the invalid URI generated?

Expl0dingBanana:
It seems that the master merge caused issues when generating routes.  I will look at this tomorrow.

Expl0dingBanana:
The issue was with that wonderful merge on GitHub.

@DrMurx / @chr2002qw  can yall validate this fixes any issues?

**WARNING** It will update your mappings file to ver 15 which migrates the mappings from URI's to ID's.  It will create a backup of the old mapping, mappings.json.14.bk which will allow you to quickly restore the old configuration.

chr2002qw:
Cannot test it currently. I pulled banana branch and MAD got stuck at "Performing Update now" since 10 minutes. I have to revert to master and fix the nginx issue with. "https://github.com/Map-A-Droid/MAD/issues/442#issuecomment-544250667" This worked for me since checkout banana branch.
Need my scanner online on Weekend because people get angry :-D

DrMurx:
Tested, there's a little issue when converting from the URI based IDs. Already discussed with @Expl0dingBanana

DrMurx:
Found another issue. `RouteManagerFactory.get_routemanager` (and through that the constructors of `RouteManagerBase` and inherited objects) used to receive the `area_id` as parameter. In the course of the changes in this PR, it accidentally became `device_id`. Therefore `RouteManagerBase.change_init_mapping` which is supposed to remove the "init" flag from an area doesn't do its job.

DrMurx:
The issue mentioned in https://github.com/Map-A-Droid/MAD/pull/446#issuecomment-546683528 is fixed with bf63718.

Issue #456, opened 2019-10-22
Please allow API to retrieve areas without specifying a mode


bbdoc:
Trying to retrieve a list of areas in order to match routemanager in trs_status table, the API currently requires a mode to be specified, which I don't know from the trs_status table. Please allow retriving the full list of areas when no mode is set on calling http://localhost:5000/api/area

Thanks

sn0opy:
Resolved via #458

PR #458, opened 2019-10-22
API - Area tweaks

Expl0dingBanana:
This PR allows /api/area to return data without requiring a mode.  If a mode is specified it will return the configuration for that mode and only areas associated with the mode.


Expl0dingBanana:
This PR is in reference to #456

sn0opy:
@bbdoc did have any time to test this?

bbdoc:
Tested and confirmed working. 

Calling http://localhost:5000/api/area now returns the full list of areas.
Calling http://localhost:5000/api/area?mode=pokestops now correctly returns only the list of areas in pokestops mode.

PR #460, opened 2019-10-22
Add injection_thresh_reboot to device settings

JabLuszko:
Looks like it got lost with new rework. Please double-check if that is all it's needed - I am not sure how this new madmin magic works :)

### DIFF
--- a/madmin/api/modules/ftr_device.py
+++ b/madmin/api/modules/ftr_device.py
@@ -280,6 +280,14 @@ class APIDevice(apiHandler.ResourceHandler):
                     "description": "Rotate accounts if player level >= 30 (for leveling mode)",
                     "expected": bool
                 }
+            },
+            "injection_thresh_reboot": {
+                 "settings": {
+                    "type": "text",
+                    "require": False,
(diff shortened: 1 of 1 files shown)


Expl0dingBanana:
Make sure to specify empty or may will cause an issue if left blank

madmin/api/modules/ftr_device.py
```diff
@@ -280,6 +280,15 @@ class APIDevice(apiHandler.ResourceHandler):
                     "description": "Rotate accounts if player level >= 30 (for leveling mode)",
                     "expected": bool
                 }
+            },
+            "injection_thresh_reboot": {
+                 "settings": {
+                    "type": "text",
+                    "require": False,
+                    "empty": None,
+                    "description": "Reboot (if enable) device after not injecting for X times in a row (Default: 20)",
```
sn0opy:
```suggestion
                    "description": "Reboot (if enabled) device after not injecting for X times in a row (Default: 20)",
```

sn0opy: @JabLuszko please approve this change to your PR. Will merge as soon as you've done so.

PR #488, opened 2019-10-30
Possibly fix dropping settings for /status and /quest page

JabLuszko:
stateDuration is set to two hours by default. Randomly (never could get pattern, those 2 hours seems like a fit) the /status page would forgot how many devices I wanted to show and it will switch back to default 10 devices. 0 means infinity.

Also fix (enable it) for /quest page. @Expl0dingBanana will give feedback, he is actively using it.

Someone please test and if the problem of /status randomly switching back to 10 devices will be fixed I will edit PR for all datatables we have there in other subpages.

### DIFF
--- a/madmin/templates/quests.html
+++ b/madmin/templates/quests.html
@@ -61,7 +61,7 @@
                 $("img.lazy").lazyload();
             },
             "autoWidth": false,
-	    "stateSave": true,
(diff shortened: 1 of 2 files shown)


Expl0dingBanana:
This change fixes the quest page issue

PR #489, opened 2019-10-30
Fixes a redirect loop when hitting the base resource

Expl0dingBanana:


Expl0dingBanana:
Accessing the page @ http://<ip_or_hostname>:5000 should now send you to the settings/devices page versus a redirect loop

acocalypso:
I tested your commit and it solves the looping issue.

PR #493, opened 2019-11-01
Jobs Page Icons

LegitDongo:
I like these icons more, and they come default with fontawesome

Example:
image

The new trash can icon will be https://fontawesome.com/icons/trash-alt?style=solid


Expl0dingBanana:
I approve of these icons (i dont think it really matters though 🤷‍♂ )

LegitDongo:
It's just for more clarity

I had no idea what the original icons meant until I hovered over them to get the tooltip title.

Issue #486, opened 2019-10-29
Automatically update quest routes by removing deleted/moved stops


sn0opy:
Kinda obvious improvement. Just wanted to have it on the todo list.

MAD should automatically update routes by

- remove deleted/moved stops
- add stops to the route that have been added/moved instead of visiting those on the 2nd run

marcb1387:
Just wanted to see if we could also log what stops get added and removed from say madmin. Some of our players try and resubmit stops that get removed and it would be nice to keep a record of that.

PR #496, opened 2019-11-01
WIP: Dead spawnpoint and stop detection

DrMurx:
This is at the very beginning of development, so expect a lot of rebasing here.

The idea is to count misses on spawns that should be there according to the database, but aren't (e.g. because it's an event spawnpoint). A "dead counter" will be increased for every miss and over a certain threshold, the spawnpoint will be considered dead. Once the spawnpoint shows a spawn again, the dead counter is reset.

Doing so will keep the gathered information about the spawnpoint's cycle time and pattern and avoids manual database operations to clean or store them after a community day or other events. A route re-calculation could then be triggered automatically, based on a threshold of dead spawnpoints, resp. on the number of spawnpoints which re-appeared.

Visually, the spawnpoints will faint on the map if their dead counter increases.

Something similar can be done for the stops (see also #486). If the stop data is missing in the MITM response, stops can be properly marked. This could automatically trigger a recalculation of the route. The Rocketmap's `pokestop` table has a field `enabled`; I don't know about its origins, but it would fit this purpose.

### DIFF
--- a/ocr/matching_trash.py
+++ b/ocr/matching_trash.py
@@ -19,7 +19,7 @@ def get_delete_item_coords(x):
     return click_x
 
 
-def trash_image_matching(screen_img):
+def trash_image_matching(screen_img, full_screen):
     clicklist: List[Trash] = []
     screen = cv2.imread(screen_img)
     # print (screen.shape[:2])
@@ -66,7 +66,10 @@ def trash_image_matching(screen_img):
         loc = np.where(res >= threshold)
         boxcount = 0
         for pt in zip(*loc[::-1]):
-            if pt[0] > width/4*3 and pt[1] < height/6*5:
+            screen_height_max = height/6*5
+            if full_screen:
+                screen_height_max = height
+            if pt[0] > width/4*3 and pt[1] < screen_height_max:
                 x_coord = int(pt[0] + tW / 2)
                 y_coord = int(pt[1] + tH / 2)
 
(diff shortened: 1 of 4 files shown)


JabLuszko:
> The Rocketmap's `pokestop` table has a field `enabled`; I don't know about its origins, but it would fit this purpose.

This is 1:1 map to POGO protos - all gyms and pokestops ("Forts") have that field set. 0 means pokestop/gym is closed (like some parks can request Niantic to disable stops at certain hours) and we can't really spin/use it. I think re-using it the way you are doing it now it's not a good idea.

sn0opy:
@DrMurx What else needs to be done for this? I am about to perform some major refactoring which will move things around and rename other things here and there.

DrMurx:
> @DrMurx What else needs to be done for this? I am about to perform some major refactoring which will move things around and rename other things here and there.

A lot, I'm afraid. I'm still working out the details. Feel free to go on with any refactoring, I'll follow your changes and rebase accordingly  - this PR is just to show that I've a grasp on that topic.

DrMurx:
The code is now on top of #548. 

I addresses the remarks from @JabLuszko and don't use the `pokestop.enabled` column, instead I've added a `dead` column to align it with the `trs_spawn`.

Still missing parts of this PR:
* Creation of those columns
* Processing incoming MITM map_proto data to find removed stops/spawnpoints. I'm currently  collecting statistical data for dead spawnpoint detection.

PR #498, opened 2019-11-01
API now processes empty values correctly

### LINKED ISSUE #495: delay_after_prio_event value when cleared is not saved
Mon-mitm mode.  Set a value for delay_after_prio_event, hit save.  Set the value to empty to clear it and save.  It retains the prior value.  To fix, one has to stop mad, edit mapping file by hand and restart mad.

Expl0dingBanana:
Fixes #495 by always processing all values received by the API.  The UI has been updated to only send fields that have changed in the current iteration


Expl0dingBanana:
lists are always sent so that will need to be changed later.  It may be better to dump the on-load data into a variable and compare with that.

PR #529, opened 2019-11-08
Skip rocketstops in levelmode

DrMurx:
Handling rocket stops during quest/level walks can cost a lot of time. Occasionally the "exit strategy" doesn't work and the account gets into the rocket fight, causing MAD to reboot the device due to a timeout waiting for proper stop data.

This pull request simply skip stops with a rocket incident in level mode. Despite such skipped stops *could* be visited safely when the incident is over, it's generally more favorable to simply go on with the planned route. Therefore, skipped rocket stops are added to the list of already visited stops (#422).

### DIFF
--- a/worker/WorkerQuests.py
+++ b/worker/WorkerQuests.py
@@ -621,12 +621,9 @@ class WorkerQuests(MITMBase):
                 else:
                     self._rocket = False
 
-                if self._level_mode and self._ignore_spinned_stops:
-                    visited: bool = fort.get("visited", False)
-                    if visited:
-                        logger.info("Levelmode: Stop already visited - skipping it")
-                        self._db_wrapper.submit_pokestop_visited(self._id, latitude, longitude)
-                        return False, True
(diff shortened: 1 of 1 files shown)


sn0opy:
@fosJoddie that's probably relevant to you since you worked on the leveling mode. Since you're already working on the next iteration, have you considered to include this as well? Can we merge this?

fosJoddie:
I added a skip for rocketized stops in #565 
When #539 gets completed I will consider to make a new option to make it configurable.

@DrMurx  : Do note that your code is now registering all rocketized stops as visited in the DB ;)

DrMurx:
@fosJoddie This can be easily mitigated. Once we hit a rocketstop in levelmode, we simply check the other stops in the GMO and if the majority (or any other threshold) is also rocketized, we do not skip.

fosJoddie:
Sure, feel free to update your PR with such a check and I feel this can be merged.

cecpk:
Not needed anymore.

PR #533, opened 2019-11-09
Fix SQL in quest scan query (potential RAM eater)

DrMurx:
Fixes a bug which makes `stop_from_db_without_quests` query all ever-seen stops without a quest, instead of just those within the bounding box around the geofence. 

This mistake doesn't bubble up to the caller because afterwards the stops are filtered to be within the fenced area, but it might contribute to higher memory consumption on databases with many pokestops.


sn0opy:
Nice catch

PR #537, opened 2019-11-10
Save team go rocket leader info if it's there

fosJoddie:
Allows us to show  team rocket leader info on pokestops


clburlison:
👍 Works perfect over here. Needs PogoDroid 1.2.1.8 with the new .eu backend.

Issue #337, opened 2019-08-18
MADmin is vulnerable to directory traversal exploits.


sebastienvercammen:
If auth is enabled, auth is required before it can be exploited (at least in this one instance I checked). If auth is disabled, _boop_ server.

**Vulnerable code:**
Below is the instance where I noticed it.

https://github.com/Map-A-Droid/MAD/blob/05042dfbf4c150dbe4465f7d29511ecbb6a6d6c7/madmin/routes/map.py#L361

```
>>> os.path.join('/a/b/c/', '../../../passwd')
'/a/b/c/../../../passwd'
```

**Steps to abuse:**
1. Use MADmin's editor to draw a new geofence.
2. Abuse the "name" field to define a system path.
3. Save.

Any path the user running MAD has access to, ending in `.txt` (since MAD appends `.txt`), can be appended.

**Possible solutions:**
- Check that a file doesn't already exist.
- Ensure the path stays within the MAD configuration folder(s).

**Other notes:**
There are more things to be done w/ `os.path.join`:
```
>>> os.path.join('/a/b/c/', '\\' + '.txt')
'\\.txt'
```
Disallowed Windows filename characters:
```
>>> os.path.join('/a/b/c/', '!!##&/\<>#*|:nope')
'/a/b/c/!!##&/\\<>#*|:nope'
```
Newlines in file names (works on Linux, can be annoying to delete if you don't already know how to):
```
>>> os.path.join('/a/b/c/', '\n')
'/a/b/c/\n'
```

The vulnerability extends to all possible uses of `open()`. I checked one other besides the first one I happened to find, and it's also vulnerable via string formatting and `origin` injection:

https://github.com/Map-A-Droid/MAD/blob/c0664c7b696f938ae4a8a30349bf0140974c4ea2/utils/adb.py#L66-L76

which gets its `origin` from a request's arguments:

https://github.com/Map-A-Droid/MAD/blob/c0664c7b696f938ae4a8a30349bf0140974c4ea2/madmin/routes/control.py#L119-L121

e.g.:
```
>>> os.path.join('/a/b/c/', 'screenshot_%s.png' % '/../../../../some-filename')
'/a/b/c/screenshot_/../../../../some-filename.png'
```

**Environment:**
* Only tested on `dev` branch

sn0opy:
I think this needs to be re-evaluated since the mappings editor has undergone som heavy changes.

Expl0dingBanana:
[...]

Issue #429, opened 2019-10-18
Move configuration to DB


sn0opy:
With the latest MADmin rework a huge milestone has been reached to move things completely into the database.

With this change a few more extras should be easier to implement like moving geofences and calculated routes to the DB and reuse geofences in multiple areas.

Expl0dingBanana:
🎉 also that looks like a taco

PR #539, opened 2019-11-11
DB migration

Expl0dingBanana:
# WORK IN PROGRESS #
### ~~*Note:  THIS WILL BREAK YOUR MADMIN / API.  DO NOT USE IN PRODUCTION*~~ This no longer breaks any functionality.

With this change all functionality from the old mappings.json file will be handled by the database.  On to the fun stuff...
* DB Schema will be created in your database
* The original mappings.json will have a slight update if you have anything with ID 0.  This conversion will be handled once and update that file.  When the ID is updated, it will update all respective elements so there should not be any loss of data.
* The configuration will be pushed into the database
* data_manager has been converted to act more as a resource factory (more on this in a bit)
* RouteManager has been updated to work with the new resources
* configmode and start have been updated to run

Resources
A new feature of the database migration are resources.  These are dict-like objects that you can interact with to easily load, set / update, and remove resources.  A very good example on how this change affected data interaction would be to refer to 'route/RouteManagerBase.py' change around init.

I am looking for any feedback where the DB conversion is not successful, the bots dont operate successfully, or you just notice something is wrong / off.


clburlison:
I duplicated my production instance and ran this PR as is, got the following traceback:
```python
[11-10 23:23:55.01] [       MainThread] [            logging:70  ] [    INFO] Setting log level to 20 (INFO).
[11-10 23:23:55.17] [         Thread-5] [      dbWrapperBase:40  ] [    INFO] Connecting to DB
[11-10 23:23:55.33] [       MainThread] [            version:28  ] [ SUCCESS] Performing update now
  File "start.py", line 170, in <module>
    version.get_version()
  File "/usr/src/app/utils/version.py", line 29, in get_version
    self.start_update()
  File "/usr/src/app/utils/version.py", line 452, in start_update
    resource = modules.MAPPINGS[section](logger, self.dbwrapper, self.instance_id)
  File "/usr/src/app/utils/data_manager/modules/resource.py", line 69, in __init__
    self.identifier = int(identifier)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

DrMurx:
My 2 cents on the table design:

* Try to get a consistent naming: primary keys should be `id` and not include the entity (which usually indicated a foreign key). `settings_monivlist`.`monlist`, `settings_walker`.`walkername`, `settings_devicepool`.`devicepool` and `settings_device`.`origin` should be simply `name`. The columns in `settings_walkerarea` don't need the prefix `walker`. I'm well aware of the origins of the column names, but we should take the opportunity to make it consistent. Since you did your own simple ORM, an additional column mapping can be added easily.
* As discussed, I would spare the overhead of having area tables split per scan-type. Most of the scan types share at least `geofence_*`, `speed` and `distance`. Having all area data in a single table makes overall handling easier, and the cost of some fields remaining NULL because they aren't used in that particular scantype can be neglected.
* I wouldn't use `ENUM` in the columns. The data we're dealing with is small so using a few more bytes for literal strings doesn't hurt, and we've full control over the application using the tables so there's no need for type safety. The disadvantage of `ENUM` is that it requires a schema change for any new value.
* Having the `monivlist_to_mon` in a relational mapping is probably too heavy, and the table currently lacks a sorting key. A VARCHAR with a list of comma separated values in `monivlist` would do it.

Expl0dingBanana:
Changes:
* Fixed the conversion issue (at least I believe I did)
* The following tables had their readable name adjusted to `name`:  settings_area, settings_device, settings_devicepool, settings_monivlist, settings_walker, settings_walkerarea
* Unique keys for instance / name have been added to the following tables: settings_area, settings_devicepool, settings_monivlist, settings_walker, settings_walkerarea
* Unique key has been added for settings_auth (index, username, password)
* Indicies have been added to settings_monivlist_to_mon
* Indicies have been added to settings_walker_to_walkerarea

"Upgrade" process (there isnt one because nothing is finalized yet)
Run the delete with foreign key check disabled
Roll back to ver16
Re-Run start or configmode

DROP TABLE `madmin_instance`, `settings_area`, `settings_area_idle`, `settings_area_iv_mitm`, `settings_area_mon_mitm`, `settings_area_pokestops`, `settings_area_raids_mitm`, `settings_auth`, `settings_device`, `settings_devicepool`, `settings_monivlist`, `settings_monivlist_to_mon`, `settings_walker`, `settings_walkerarea`, `settings_walker_to_walkerarea`;

Expl0dingBanana:
You should now be able to access any page in madmin and not receive any error messages.  Creating, updating, and deleting is all done through the API which has not been updated yet.

Expl0dingBanana:
This should be good for mass-testing as it has feature parity from master.

Expl0dingBanana:
I guess I should link the DB Migration issue #429 so it can be closed whenever this is merged.

Expl0dingBanana:
Things that might be beneficial prior to merge:
* Move GeoFences to DB
* Move Route Calcs to DB
* Move Images to DB
* Move PlayerStats to DB

After moving these, that would fix #337 since there are no more file opening / paths exposed

db/madmin_conversion.py
```diff
@@ -0,0 +1,306 @@
...
+    ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;""",
+
+    """CREATE TABLE IF NOT EXISTS `settings_auth` (
+        `auth_id` int(11) NOT NULL AUTO_INCREMENT,
+        `guid` varchar(32) NULL,
+        `instance_id` int(11) NOT NULL,
+        `username` varchar(32) COLLATE utf8mb4_unicode_ci NOT NULL,
+        `password` varchar(32) COLLATE utf8mb4_unicode_ci NOT NULL,
+        PRIMARY KEY (`auth_id`),
+        -- UNIQUE (`instance_id`, `username`, `password`),
```
DrMurx: This unique key is commented out. 

I would also not do a unique over these 3 fields but just over `instance_id` and `username`.

db/madmin_conversion.py
```diff
@@ -0,0 +1,306 @@
+TABLES = [
+    """CREATE TABLE IF NOT EXISTS `madmin_instance` (
+        `instance_id` INT NOT NULL AUTO_INCREMENT,
+        `name` VARCHAR(128) NOT NULL,
+        PRIMARY KEY (`instance_id`),
+        UNIQUE KEY (`name`, `instance_id`)
```
DrMurx: This unique should be `name` only.

db/madmin_conversion.py
```diff
@@ -0,0 +1,306 @@
+TABLES = [
+    """CREATE TABLE IF NOT EXISTS `madmin_instance` (
+        `instance_id` INT NOT NULL AUTO_INCREMENT,
```
DrMurx: Consider having all primary and foreign keys `int unsigned`.

db/madmin_conversion.py
```diff
@@ -0,0 +1,306 @@
...
+
+    """CREATE TABLE IF NOT EXISTS `settings_walkerarea` (
+        `walkerarea_id` int(11) NOT NULL AUTO_INCREMENT,
+        `guid` varchar(32) NULL,
+        `instance_id` int(11) NOT NULL,
+        `area_id` int(11) NOT NULL,
+        `algo_type` enum('countdown','timer','round','period','coords', 'idle') COLLATE utf8mb4_unicode_ci NOT NULL,
+        `algo_value` varchar(256) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+        `max_walkers` int(11) DEFAULT NULL,
+        `name` varchar(128) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
```
DrMurx: For consistency, `name` should be on the 4th position.

utils/data_manager/modules/walker.py
```diff
@@ -0,0 +1,94 @@
...
+        super().save(core_data=core_data, force_insert=force_insert)
+        # Get all current walkerareas
+        sql = "SELECT `walkerarea_id` FROM `settings_walker_to_walkerarea` WHERE `walker_id` = %s"
+        walkerareas = self._dbc.autofetch_column(sql, args=(self.identifier,))
+        removed_walkerareas = set(walkerareas) - set(self._data['fields']['setup'])
+        # Remove old walkerareas from the table
+        del_data = {
+            'walker_id': self.identifier
+        }
+        self._dbc.autoexec_delete('settings_walker_to_walkerarea', del_data)
```
DrMurx: This is a potential threat to data safety. If anything happens between deletion and insertion, the configuration might be destroyed. 

The easiest way to fix this is to wrap the `save` methods in a database transaction, commit on success and rollback on exception. This should not happen within each `save` method (as they might be nested). Instead, create a `save` method on the base class to do what all model classes have in common (transaction, presave validation etc), and an abstract `_save` which does the per-model work.

db/madmin_conversion.py
Expl0dingBanana: I am on the fence about these unique keys for MADmin settings.  They are the correct way to go as they can cause unintended consequences.  However its how the current config works and to my knowledge no one has complained so far.  I think we need input from @sn0opy or @cecpk to figure out the route they would like us to take.

db/madmin_conversion.py
Expl0dingBanana: Agreed.  Resolved in 46ad21a

db/madmin_conversion.py
Expl0dingBanana: Agreed.  Resolved in 46ad21a

db/madmin_conversion.py
Expl0dingBanana: Agreed.  Resolved in 46ad21a

utils/data_manager/modules/walker.py
Expl0dingBanana: Ill need to do some extra work to allow transactions as the current implementation of execute is only good for a single query.  I think using __enter__ and __exit__ would be a good way to retrieve a cursor for multiple queries (start transaction, queries, commit).

db/madmin_conversion.py
DrMurx: Eventually it wouldn't matter to have multiple accounts with the same name and different passwords. It appears just "uncommon" in general...

Expl0dingBanana:
This is all of the changes I would like to make to this PR.  I will add unit-tests for geofences and routecalcs, and fix any unit-test that uses areas.  The area unit-tests need to be redesigned as fields have changed in the migration.

Please let me know if anyone encounters any issues.

clburlison:
After countless DMs to @Expl0dingBanana this 100% converts my prod instance successfully. Huge thanks for the work, I know it wasn't a simple addition.

Notes:
* The template for the "Devices" page looks quite bad on mobile with the bulk update addition. (See attachment below) 
* Setting the Geofence name to `foobar` instead of `configs/geofences/foobar.txt` would be a welcome addition however it's quite easy to update post upgrade so 🤷‍♂ 
* I still would prefer the monivlist to be a json blob stored in a TEXT field instead of separated out into two tables but I can script the API for my needs, also 🤷‍♂
* madmin UI is slightly slower to load a few pages due to the additional database calls but acceptable. Talking less than 2 seconds on heavy pages compared to the previous json config which was in the ~.5 second range.
* Bulk update devices to a new walker is 👍 👍 👍 👍 

<details>
  <summary>Click to expand image</summary>
IMG_4394
</details>

bbdoc:
`Move PlayerStats to DB Probably not worth it`

Wondering why this was taken out. Probably not worth it in terms of file access indeed, but at the same time, we already have a trs_status holding player info, so adding a few columns to this table and storing player stats in it would be pretty easy.

DrMurx:
> we already have a trs_status holding player info, so adding a few columns to this table and storing player stats in it would be pretty easy

I would argue that PlayerStats should be stored in a dedicated table `trs_pogo_accounts`. It's not device related data, but account related. This can be done in another iteration of the DB migration, no need to include it here.

Expl0dingBanana:
Looks like when I had updated the required fields (f048a2a) I forgot to update the DB schema.  Please run the following commands:

ALTER TABLE `settings_devicepool` CHANGE `screenshot_type` `screenshot_type` ENUM('jpeg','png') CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NULL;
ALTER TABLE `settings_area_pokestops` CHANGE `level` `level` TINYINT(1) NULL;
ALTER TABLE `settings_area_raids_mitm` CHANGE `including_stops` `including_stops` TINYINT(1) NULL;
ALTER TABLE `settings_device` CHANGE `screenshot_type` `screenshot_type` ENUM('jpeg','png') CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NULL;
ALTER TABLE `settings_devicepool` CHANGE `screenshot_type` `screenshot_type` ENUM('jpeg','png') CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NULL;

sn0opy:
Talked to @Expl0dingBanana. This seems to be basically production ready. At least those who tested this are happy and didn't report any other issues so far.

PR #543, opened 2019-11-11
fix egg moveset

muckelba:
as @fosJoddie said in discord, egg moves should be 0


sn0opy:
Keeping this open a bit longer. I'll look into MAD's code, so we will probably change this to `NULL` instead.

sn0opy:
Going to close this for now but will work on this and actually set it correctly to NULL including checks at several places.

PR #552, opened 2019-11-15
Fix Decimal interpretation ending in TypeError - so cast to float

florianbecker:
Fixing type mismatching, maybe from mysql client. This will bringt back the mons statistic working with mysql 8


sn0opy:
Nice find!

PR #553, opened 2019-11-16
Sort screentypes and detect bans

stonedDiscord:
I sorted the screentypes after their numbers.
Added OCR texts to detect banned accounts.
Added detection of a warning strike by comparing the most present color which is 16,24,33 when the strike is displayed.
Returning player screen comes after selecting the birthdate.
Account type selection comes after pressing "returning player"

### DIFF
--- a/ocr/pogoWindows.py
+++ b/ocr/pogoWindows.py
@@ -32,6 +32,7 @@ class PogoWindows:
 
         self._ScreenType: dict = {}
 
+        self._ScreenType[1]: list = ['Geburtdatum', 'birth.', 'naissance.', 'date']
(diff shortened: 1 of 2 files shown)


sn0opy:
What happens if your device got stuck on an all-black loading screen, or you forgot to disable battery saving mode? I mean, in general it looks like it just throws warnings and does not actually do anything with that information, right?

stonedDiscord:
True, all it does is throw a warning and continues.
I suggest that we do a vote on how to handle the bans.
On the colors:
grafik
left is all black while loading, right is the warning bg.
It's not actually black but more like a dark blue, so i assume that it will detect the difference, thus meaning that nothing will change when the screen is all black (0,0,0) while loading.

PR #555, opened 2019-11-16
add items to the DO NOT DELETE list

stonedDiscord:
There now is an additional list with items that can be safely deleted. (Pokeballs, Potions, Berries)
Gifts can now be deleted by the bot.
Mysterious components
Raid Passes
Lures
Star Pieces
Evolution items
are now all exempt from deletion.

### DIFF
--- a/worker/WorkerQuests.py
+++ b/worker/WorkerQuests.py
@@ -440,6 +440,7 @@ class WorkerQuests(MITMBase):
                     'Nanab Berry', 'Nanabbeere', 'Baie Nanab',
                     'Pinap Berry','Sananabeere', 'Baie Nanana')
         not_allow = ('Lucky Egg', 'Glücks-Ei', 'Glucks-Ei', 'Gliicks-Ei', 'CEuf Chance', 'Œuf Chance',
+                    'Gift', 'Geschenk', 'Cadeau',
                     'Camera', 'Kamera', 'Appareil photo',
                     'Mystery Box', 'Wunderbox', 'Boîte Mystère',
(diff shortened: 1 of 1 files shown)


ChaosEmpire:
Why you want to delete 10 gits, or 6 pieces (which you only get after fighting rockets)

it means an iteration of ocr for a handful items
Better just delete those with huge amoutn in one step and you are much faster

i changed my button press time so it deletes ~ 150 pokeballs in one step, , result 10 stops per device more in 1 hour

I would be happy if those postions with just a handful of items would get ignored before stepping into deletion routine

JabLuszko:
> Gifts can now be deleted by the bot.

You tested that, right? Gifts has totally different deleting menu, I was pretty sure that MAD code never could those (because waste of time and OCR)?

stonedDiscord:
i've set the gifts to no-delete because the menu really is complicated for a bot and i didn't check that.
i also doubled the time that the bot holds the + button so it gets rid of as many items as it can in one stretch. Before i did have the situation that it didn't hold the button for long enough and had to delete the same item twice, even when it had less than 50 of that item.

domizei385:
I’ve had bots with e.g. 20+ dragon skins blocking the inventory.
I think those should be cleared as well, just less frequently than poke balls. Maybe with a 5% chance?!

stonedDiscord:
it is probably better to make this configurable via the db/madmin

Expl0dingBanana:
A system page will be added with #657. It's on my to-do list to make all of this easily configurable through madmin

PR #565, opened 2019-11-26
Refactoring of quest level mode

fosJoddie:
Should make quest level mode be more memory friendly, especially for large routes.

Changes:

- Each worker will get their own list based on what has been visited before
- A worker will use the saved routecalc as long as there are more or equal unvisited stops in db
- If there's fewer unvisited than the routecalc a recalc of route WILL happen. This might have negative impact on very large routes with a low number of stops in the routefile. Make sure you're on quick calc mode for those cases.

**Note**
- There's only one list of stops considered to ignored. This means if more than one worker is in the area and one worker fails the spinnable check, it will be added to the ignore list for everyone.
- Whenever a worker joins an area, all worker routes will be recalculated.

### DIFF
--- a/db/DbWrapper.py
+++ b/db/DbWrapper.py
@@ -8,6 +8,7 @@ from functools import reduce
 import mysql
 from bitstring import BitArray
 
+from geofence.geofenceHelper import GeofenceHelper
 from utils.collections import Location, LocationWithVisits
 from utils.logging import logger
 from utils.s2Helper import S2Helper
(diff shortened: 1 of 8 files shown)


sn0opy:
#548 has just been merged, please make sure to adapt to certain changes.

fosJoddie:
Rebased on master and moved my new db methods. Thanks for the heads up @sn0opy  👍

bbdoc:
Not sure what you mean with `If there's fewer unvisited than the routecalc a recalc of route WILL happen`. I have a routecalc on Paris with 12k stops. If I launch my instance 10 mins and then need to restart it, it will already have less unvisited than routecalc so it will recalculate a route on nearly 12k stops... pretty inefficient even if using quick calc. Or did I misunderstood something ?

If that's the way then would really be nice to have an option to deactivate this and simply keep the existing route minus visited stops (current behaviour).

fosJoddie:
@bbdoc 
> If there's fewer unvisited than the routecalc a recalc of route WILL happen. This might have negative impact on very large routes with a low number of stops in the routefile. Make sure you're on quick calc mode for those cases.

Yeah this was a bad design on my part. Changed it to:
* As long as a route from original route can be used, it will be used

PR #567, opened 2019-11-28
Fix "Last mon seen" in MAdmin

Seandals:
Currently on the map in MAdmin the "Last mon seen" is showing the last confirmation (last_non_scanned) even if MAD has seen a spawn more recently (last_scanned). Variables have been swapped. Better explained in my Discord screenshot below.

discordexplanation

### DIFF
--- a/madmin/static/js/madmin.js
+++ b/madmin/static/js/madmin.js
@@ -1198,9 +1198,9 @@ new Vue({
       var last_non_scanned = moment(spawn["lastnonscan"])
 
       if (last_scanned.isBefore(last_non_scanned)) {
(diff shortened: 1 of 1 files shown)


Seandals:
After a few hours of sleep, I came back to my map and realized that swapping the vars had produced the unintended consequence of breaking the "Last mon seen" on spawnpoints where the timer had not yet been figured out.

I toyed around with it for a while when I realized that the before logic really didn't even need to be there at all and feeding the lastnonscan directly into the html produces the correct result in the map.

Before:
beforeactive
beforeunknown

After:
afteractive
afterunknown

PR #585, opened 2019-12-06
Attempt to disable vibration when starting pogo

MonsterMap:
This seemed to be the best location to put this, as it only needs to be done once per boot.
Tested working on my devices, will just silently fail on devices that don't vibrate (TV boxes).


simonbuehler:
💤 👍

PR #598, opened 2019-12-09
Fix for #597

### LINKED ISSUE #597: Max_Devices in Walker Settings does no longer work
I have a walker configured with several areas, one of them is set to max devices 0 (cause i only activate it for cdays but max device 0 also get used a lot in old level mode)

Instead of skipping this area like before the db migration, all devices move into it and do the wrong work

I changed the setting to 2, databse is correct, but still all devices run into this route, instead just 2 in that case
------
Btw in Database it s max_walkers, in configuration option its called max walkers in area
the coloumn of the setting is called Max Devices, maxbe it should also get renamed to Max Walker

Expl0dingBanana:
Fixes #597

### DIFF
--- a/worker/WorkerBase.py
+++ b/worker/WorkerBase.py
@@ -124,7 +124,7 @@ class WorkerBase(ABC):
 
     def check_max_walkers_reached(self):
         walkermax = self._walker.get('walkermax', False)
-        if not walkermax:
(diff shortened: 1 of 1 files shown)


ChaosEmpire:
Tested to be working, #597 can be closed when merged

PR #605, opened 2019-12-10
Updated trs_status.instance to int

Expl0dingBanana:
trs_status.instance is being removed and converted to trs_status.instance_id to avoid collation errors across tables

### DIFF
--- a/db/DbSchemaUpdater.py
+++ b/db/DbSchemaUpdater.py
@@ -186,9 +186,9 @@ class DbSchemaUpdater:
         },
         {
             "table": "trs_status",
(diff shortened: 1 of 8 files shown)


sn0opy:
Eventhough I manually updated my tables already, this does seem to work without any issues. Still looking forward to get more feeback on this before merge

hrueppert:
I tested it, even with an OSM Update and run build everything looks fine.
Table collation is correct and stays correct. Status Page and Quest Page work perfectly.

kettnerflorian:
It works well for me, too. No further problems have occurred so far.

PR #607, opened 2019-12-11
QuestRoute Fix

Expl0dingBanana:
Quest routes were not being regenerated to pickup new stops

### DIFF
--- a/utils/data_manager/modules/routecalc.py
+++ b/utils/data_manager/modules/routecalc.py
@@ -62,21 +62,23 @@ class RouteCalc(resource.Resource):
     # =====================================================
 
     def calculate_new_route(self, coords, max_radius, max_coords_within_radius, delete_old_route, calc_type,
-                            useS2, S2level, num_procs=0, overwrite_calculation=False):
+                            useS2, S2level, num_procs=0, overwrite_calculation=False, in_memory=False):
         if overwrite_calculation:
             calc_type = 'quick'
-        if delete_old_route:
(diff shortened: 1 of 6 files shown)


utils/data_manager/modules/routecalc.py
```diff
@@ -62,21 +62,23 @@ def save(self, force_insert=False, ignore_issues=[]):
     # =====================================================
 
     def calculate_new_route(self, coords, max_radius, max_coords_within_radius, delete_old_route, calc_type,
-                            useS2, S2level, num_procs=0, overwrite_calculation=False):
+                            useS2, S2level, num_procs=0, overwrite_calculation=False, in_memory=False):
         if overwrite_calculation:
             calc_type = 'quick'
-        if delete_old_route:
+        if delete_old_route and in_memory is False:
```
fosJoddie: Does this not save the empty route to db?

Expl0dingBanana: The empty route is saved a few lines below and is executed from This section will be executed by https://github.com/Map-A-Droid/MAD/pull/607/commits/7537beaad9cfe68e510c6c7f28843a573137bf0c#diff-cb3f86e4a1cf1897e04358904eed82e6R35

fosJoddie: Let me rephrase:  why do we overwrite the route in db to redo a route in-memory?

fosJoddie: 😬 disregard. I read it as If true 🙈

Issue #596, opened 2019-12-09
DbWrapper.py retrieve_next_spawns can't retrieve 30 minute spawns between xx:20 and xx:30 each hour


crhbetz:
After observing that my mon_mitm prioQ gets smaller and smaller whenever approaching half of the hour (:30) I looked into the db query used to get the spawn info for prioQ: https://github.com/Map-A-Droid/MAD/blob/ebcc727724c1ad13de133d7cd2dd9717e05830bb/db/DbWrapper.py#L788

Between xx:20 and xx:30 the endtimes of 30 minute spawns (spawndef != 15) cross the full hour mark and I suspect that the BETWEEN part of the query doesn't work in this situation (for example at xx:25, it would become `BETWEEN 55:00 and 05:00`).
The same behaviour occurs for 60 minute spawns (spawndef = 15) between xx:50 and xx:59.

I think this would cause a significant amount of spawn points to never get scheduled in prioQ mode (1/6?).

Unfortunately, I have no idea how to solve this issue.

Below are sample queries getting counts, first set only getting 30 minute spawns at 14:50:59, second set only getting 60 minute spawns at 14:20:59 (note the added `AND spawndef != 15` vs `AND spawndef = 15` to point out the difference between 30 and 60 minute spawns):

```
MariaDB []> SELECT count(*) FROM trs_spawn WHERE calc_endminsec IS NOT NULL AND (latitude >= 1 AND longitude >= 1 AND latitude <= 99 AND longitude <= 99) AND DATE_FORMAT(STR_TO_DATE(calc_endminsec,'%i:%s'),'%i:%s') BETWEEN DATE_F
ORMAT(DATE_ADD("2019-12-09 14:50:59", INTERVAL if(spawndef=15,60,30) MINUTE),'%i:%s') AND DATE_FORMAT(DATE_ADD("2019-12-09 14:50:59", INTERVAL if(spawndef=15,70,40) MINUTE),'%i:%s') AND spawndef != 15;     
+----------+
| count(*) |
+----------+
|      438 |
+----------+
1 row in set (0.01 sec) 
          
MariaDB []> SELECT count(*) FROM trs_spawn WHERE calc_endminsec IS NOT NULL AND (latitude >= 1 AND longitude >= 1 AND latitude <= 99 AND longitude <= 99) AND DATE_FORMAT(STR_TO_DATE(calc_endminsec,'%i:%s'),'%i:%s') BETWEEN DATE_F
[...]

Issue #579, opened 2019-12-05
Existing PrioQ gets dropped on every recalculation (mon_mitm)


crhbetz:
Bringing this over from discord for better readability and to keep organized.
I only looked at **mon_mitm** for this so far.
I'm not creating a PR because I'm not convinced I understand everything going on here and want to post it for discussion.

In `_update_priority_queue_loop`, `_merge_priority_queue` is used (L320), which returns a new (clustered) prioQ that doesn't take the existing prioQ in `self._prio_queue` into account, instead just replacing it.

https://github.com/Map-A-Droid/MAD/blob/20b8b0eb6477fb106d0b5422cc7640e17b8491b6/route/RouteManagerBase.py#L313-L327

https://github.com/Map-A-Droid/MAD/blob/20b8b0eb6477fb106d0b5422cc7640e17b8491b6/route/RouteManagerBase.py#L329-L338

`_update_priority_queue_loop` runs every 180 seconds in mon_mitm mode:

https://github.com/Map-A-Droid/MAD/blob/20b8b0eb6477fb106d0b5422cc7640e17b8491b6/route/RouteManagerMon.py#L6-L7

New Q items for mon_mitm are fetched through `retrieve_next_spawns` in dbwrapper.py.
I don't understand the database query there, but the function excludes spawns in the past, even explicitly stating so in the comment:
https://github.com/Map-A-Droid/MAD/blob/20b8b0eb6477fb106d0b5422cc7640e17b8491b6/db/DbWrapper.py#L797-L800

This results in the prioQ never containing spawns that are older than ~4 minutes. I think this is both unexpected behaviour and probably not intended, because the `remove_from_queue_backlog` would only do anything if set below ~3-4 minutes, which is way too aggressive.

The behaviour of discarding the previous prioQ was introduced in https://github.com/Map-A-Droid/MAD/commit/328e807fae87bcda212250fc4f5903c388e53b9f - apparently to fix issues of the prioQ getting too large.
I'd guess this problem would also be solved by setting either a reasonable default value for `remove_from_queue_backlog` (not 0) or a reasonable fallback value here:
https://github.com/Map-A-Droid/MAD/blob/328e807fae87bcda212250fc4f5903c388e53b9f/route/RouteManagerBase.py#L326-L327

I propose the following diff to try out and analyze reverting the change from April:
```diff
diff --git a/route/RouteManagerBase.py b/route/RouteManagerBase.py
index 94762b3..34cff2f 100644
[...]

PR #610, opened 2019-12-12
prioQ fixes for mon_mitm, little improvements, logging

crhbetz:
fixes https://github.com/Map-A-Droid/MAD/issues/579 by not dropping the prio q on recalculation for mon_mitm, but also reducing recalculation to one per 59 minutes,
fixes https://github.com/Map-A-Droid/MAD/issues/596 by ditching the complicated database query and just fetching all spawns for the next hour instead.

Both of these changes are feasible, because once all spawn / despawn times are known, the prioQ for mon_mitm actually just becomes a line sorting the spawn points by their spawn times. No actual prioQ flexibility is needed in this mode. It should lead to more robust and humanly understandable scanning behaviour. Also, I added a little code to check each prioQ event against the `remove_from_queue_backlog` threshold, because otherwise this would only happen during prioQ recalculation. It makes more sense this way imo, but it is also needed with hourly recalculations only, or the prioQ might fall very far behind. If an event is over the threshold, a loglevel warning message will be logged.

I limited changes of behaviour to mon_mitm mode for now because it might have unintended side effects on other modes.

I also added a couple of log lines and a little detail to existing lines to make it more understandable what is happening with prioQ.

### DIFF
--- a/db/DbWrapper.py
+++ b/db/DbWrapper.py
@@ -802,9 +802,6 @@ class DbWrapper:
             "FROM trs_spawn "
             "WHERE calc_endminsec IS NOT NULL "
             "AND (latitude >= {} AND longitude >= {} AND latitude <= {} AND longitude <= {}) "
-            "AND DATE_FORMAT(STR_TO_DATE(calc_endminsec,'%i:%s'),'%i:%s') BETWEEN DATE_FORMAT(DATE_ADD(NOW(), "
-            " INTERVAL if(spawndef=15,60,30) MINUTE),'%i:%s') "
-            "AND DATE_FORMAT(DATE_ADD(NOW(), INTERVAL if(spawndef=15,70,40) MINUTE),'%i:%s')"
         ).format(minLat, minLon, maxLat, maxLon)
 
         res = self.execute(query)
--- a/db/madmin_conversion.py
+++ b/db/madmin_conversion.py
@@ -14,6 +14,11 @@ COLUMNS = [
         "column": "last_updated",
         "ctype": "DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP AFTER `recalc_status`;"
     },
+       {
+        "table": "settings_area_mon_mitm",
+        "column": "max_clustering",
+        "ctype": "int DEFAULT NULL"
+    },
 ]
 TABLES = [
     """CREATE TABLE IF NOT EXISTS `madmin_instance` (
--- a/route/RouteManagerBase.py
+++ b/route/RouteManagerBase.py
@@ -247,8 +247,15 @@ class RouteManagerBase(ABC):
                 self._stop_update_thread.clear()
             self._prio_queue = []
             if self.mode not in ["iv_mitm", "pokestops"]:
+                if self.mode == "mon_mitm" and self.settings is not None:
+                    max_clustering = self.settings.get(
+                        "max_clustering", self._max_coords_within_radius)
+                    if max_clustering == 0:
+                        max_clustering = self._max_coords_within_radius
+                else:
+                    max_clustering = self._max_coords_within_radius
                 self.clustering_helper = ClusteringHelper(self._max_radius,
-                                                          self._max_coords_within_radius,
+                                                          max_clustering,
                                                           self._cluster_priority_queue_criteria())
             self._update_prio_queue_thread = Thread(name="prio_queue_update_" + self.name,
                                                     target=self._update_priority_queue_loop)
@@ -503,7 +510,7 @@ class RouteManagerBase(ABC):
         delete_seconds_passed = 0
         if self.settings is not None:
             delete_seconds_passed = self.settings.get(
-                "remove_from_queue_backlog", 0)
+                "remove_from_queue_backlog", 900)
 
         if delete_seconds_passed is not None:
             delete_before = time.time() - delete_seconds_passed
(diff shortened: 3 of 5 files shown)


sn0opy:
Could you give us a short explanation what to look for when testing this?

crhbetz:
It's a bit tricky, because right now there's no comparable logging at the points where I made changes.

There should be an increase in scanning performance especially because of the fix of #596 , but it would still be hard to read from statistics because when the prioQ would get few events, the workers would resume walking the normal route and still scan stuff. It should provide better / more stable time left on scanned mons especially leading up to the half hour mark, but that's again difficult to track.

The best you can probably do is compare the prioQ on the map (or the printed one from loglevel debug) with the spawn times of the points it wants to check, to see if the Q remains consistent - especially to confirm everything going right across the half hour mark and change of hours.

One might set a really aggressive `remove_from_queue_backlog` for a bit to see if that new part correctly skips overdue events.

ChaosEmpire:
Remove any events in priority queue that have been due for scanning before NOW - given time in seconds (Default: 0)

I understand that default 0 means , disable removing of entries in backlog
but i get lots of remove backlog entries

assuming that 300 means, remove anything older than 300 seconds from backlog, so a 30min spawn can max get find as 25min its possible that 0 means reove directly all backlog, but then we need to set high values to workaround removal of backlog entries, so it tries to do all work

btw the part min_time_left_seconds could be used in similiar way for the backlog
if we dont want to report pokemon with less than x seconds..we also dont plan a route for them and can remove them from backlog, even if backlog removal is disabled

from my understanding backlog removal makes only sense in a range of 1sec after event to min_time_left_seconds value

maybe this handling can be checked too

ChaosEmpire:
During a round based mon mitm, a 2nd device has the chance to encounter iv missing pokemon
how about using "idle" times to handle not encountered pokemon in a 2nd hop
something like mon mitm device pool does interact like an iv mitm scanner device and cleans the iv mitm queue that way
if you have enough devices, you will have sometimes a few seconds without a jump, in this moment a device could encounter pokeon nr3/4 of a cluster, this will increase the amount of iv scanned pokemon, instead of let the devices have "idle" moments, this could make a only iv mitm scanner obsolete and device usage more efficient, as sometimes iv mitm devices stand around for minutes
also it would increase the max amount of pokemon on the map, if all device could be used as a combined mon mitm/iv mitm device

crhbetz:
Thanks for your input @ChaosEmpire - you brought up some valid points.

1. regarding `min_time_left_seconds` - as I just confirmed in the code, this will only be used in `iv_mitm` mode at all. Additionally, the prioQ build process in `mon_mitm` only fetches future spawn events from the database, so the value would be irrelevant anyway. (after 1 hour when the next calculation happens, older events are not automatically discarded - the newly built Q will be appended and the older events would be handled by `remove_from_queue_backlog` if necessary).

2. regarding remove_from_queue_backlog: `assuming that 300 means, remove anything older than 300 seconds from backlog, so a 30min spawn can max get find as 25min its possible that 0 means reove directly all backlog, but then we need to set high values to workaround removal of backlog entries, so it tries to do all work`
You're correct that with my changes, the default value of 0 makes no sense, because it will indeed remove any event that's older than "now". However, you probably don't want to disable removing old events completely, because in situations like partial outages of your devices you could end up with a Q that's behind hours and hours and in essence jumping around aimlessly.
I think the proper way to handle it would be a sane default value like 900 seconds (15 minutes). I'd have to check if this would negatively / counter-intuitively impact other scan modes, though. Also, the description of `remove_from_queue_backlog` could be improved.
I will check this and submit a change regarding this.

3. mon_mitm + iv_mitm "collaboration" / "merge"
I understand your proposal and where you identify the possible advantage, however this is beyond both the scope of this PR and my current grasp of the code base. I think this could only be achieved in a new scan mode / a merge of mon_mitm and iv_mitm.
Further, if your mon_mitm workers already have downtimes, I think the more appropriate solution would be to limit event clustering to a maximum of 2 events (something I've already thought about making configurable too), so that when running smooth, there should never be more than 2 encounters available on any given event, thus "guaranteeing" a 100% encounter rate.
I will think about and look into the latter option.

crhbetz:
@ChaosEmpire 🏓 

- Added `max_clustering` option to `mon_mitm` settings. Most probable use case would be to set it to `2`, so that a maximum of two mon events get clustered and thus in theory - provided your devices can keep up at all times! - all mons should get IV checked (catering to the two mons per location / GMO limitation of PogoDroid). Default is no limit / the value of `max_coords_within_radius`.
Search for the line containing `clustered new events` in logs to see the result of this. On my areas, I get an increase of 30-35% in clustered events from setting this to two.
- removed `min_time_left_seconds` from `mon_mitm` settings (not removed from DB yet)
- `remove_from_queue_backlog` now defaults to `900`, reasoning is detailed in my previous comment

ChaosEmpire:
I tested the PR based on max clustering unlimited, 2 and 3 for several areas i m running in mon mitm mode

one area is surley in a very good relation of device amount versus spawns
the others are mixed to very bad density of devices

result of testing, lots of more pokemon on the map compared to original prioq, 
times surely better compared to mode without prioq, but with cluster 2 its nearly the same of amount of pokemon and iv scanned, with max cluster 3 more pokemon, without big change of iv scanned amount, but still with much better times, than round based, so for the first time after month i would say, its worth to use prioq
the area with few devices and much more spawnpoints profits the most, amount of pokemon increased between 60 and 80 percent, , times got better, and iv checked pokemon increaed,compared to round based and also old prioq was less good than this

From my point of view, better logging, better scanning...
good pr, since last changes more customizable

Seandals:
I have also been testing this PR for the past several days and have noticed a significant improvement using prioq in a large area with a large number of spawn points. No issues. This is an awesome update!

PR #613, opened 2019-12-13
fix routecalc so it doesnt save in memory mode

fosJoddie:


sn0opy:
2 People confirmed this fix is working.

PR #615, opened 2019-12-14
UX for route recalulation

Expl0dingBanana:
During route recalcuation the UI now displays an icon showing that it is being being recalculated.  The 'areas' page now checks the status every 5 seconds to determine the status of all recalcuatlions fro all areas.  If the route is being recalculated it cannot be triggered again.

The schema for routecalcs has been updated to include the status (recalc_status) which states if the recalculation is running and when the last time the route was re-calculated.  These values are reset on boot since any previously running thread has been ended

### DIFF
--- a/configmode.py
+++ b/configmode.py
@@ -44,7 +44,7 @@ if __name__ == "__main__":
 
     instance_id = db_wrapper.get_instance_id()
     data_manager = utils.data_manager.DataManager(db_wrapper, instance_id)
-    db_wrapper, db_pool_manager = DbFactory.get_wrapper(args)
+    data_manager.clear_on_boot()
     version = MADVersion(args, data_manager)
     version.get_version()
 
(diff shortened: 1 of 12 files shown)


Expl0dingBanana:
The icons are loaded through an AJAX call so you will see them appear after the page loads.  Expected loaded result shows the blue recalculate (refresh) icon.
image

While the recalculation is progress the typical MAD loading icon should replace the recalculation button
image

The page pulls the status every 5 seconds.  Once the calculation is done (and the 5s interval refresh is triggered, you should see the result from the load results.

Expl0dingBanana:
For testing the API you can force the recalculation with the following curl command:
`curl -v -H 'X-Beautify: 1' -X POST -H 'Content-Type: application/json-rpc' -d '{"call": "recalculate"}' http://localhost:5000/api/area/<area_id>`

If the recalculation is not running, you will receive a HTTP/204 status code.  If the recalculation is running a HTTP/422 will be returned.

JimBim89:
Maybe disable icons for iv_mitm and idle modes since they are routeless?

PR #616, opened 2019-12-14
New instance startup fixes

Expl0dingBanana:
This change allows for new instances to successfully start from a base installation (db + OSM-RM requirements).  The old mappings file is no longer required to successfully startup

Tested on a new installation (blank db -> import osm-rocketmap requirements), a v14 mappings.json file -> v19.

### DIFF
--- a/configmode.py
+++ b/configmode.py
@@ -37,10 +37,6 @@ if __name__ == "__main__":
             'config.ini file not found. Check configs folder and copy example config')
         sys.exit(1)
 
-    filename = args.mappings
-    if not os.path.exists(filename):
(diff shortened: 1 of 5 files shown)


Expl0dingBanana:
I will be adding the multi-instance conversion fixes to this PR.

Expl0dingBanana:
For the multi-instance test run the migration on your main instance.  Review the output around for any issues.  Update config.ini to include status-name (or update to a new name), roll back version.json to v16, and re-run the conversion.  Validate that any issues you saw from the initial conversion are present and that no new issues were created.  Look at the DB and validate that all entries are in both instances with different IDs.

PR #617, opened 2019-12-15
New Instance Setup + Multi-Instance Conversion

Expl0dingBanana:
This PR allows for a new system to be setup from scratch and fixes several issues where multi-instance MAD instances were not converting correctly.

To test the multi-instance conversion:
1)  Convert to DB
2)  Edit config.ini and add status-name and set it to a value other than mad.  If this is already set, change the status name
3)  Roll back version.json to 16.
4)  Re-Run the upgrade.

To clear out from a multi-instance conversion you can run the following queries (where instance_id is from madmin_instance)
DELETE FROM settings_device WHERE instance_id = <instance_id>;
DELETE FROM settings_walker_to_walkerarea WHERE walker_id IN (SELECT walker_id FROM settings_walker WHERE instance_id = <instance_id>);
DELETE FROM settings_walker_to_walkerarea WHERE walkerarea_id IN (SELECT walkerarea_id FROM settings_walkerarea WHERE instance_id = <instance_id>);
DELETE FROM settings_walker WHERE instance_id = <instance_id>;
DELETE FROM settings_walkerarea WHERE instance_id = <instance_id>;
DELETE FROM settings_area WHERE instance_id = <instance_id>;
DELETE FROM settings_monivlist WHERE instance_id = <instance_id>;
DELETE FROM settings_routecalc WHERE instance_id = <instance_id>;
DELETE FROM settings_geofence WHERE instance_id = <instance_id>;
DELETE FROM settings_devicepool WHERE instance_id = <instance_id>;
DELETE FROM settings_auth WHERE instance_id = <instance_id>;

Issue #599, opened 2019-12-09
Idle Device not shown in Status


ChaosEmpire:
I have a walker set to idle mode with area in idle
Devices added to this idle walker have been displayed in Status Page before , after db migration they dont appear any longer

Expl0dingBanana:
Are you still seeing this issue in the latest master?

PR #619, opened 2019-12-17
QoL updates

Expl0dingBanana:
- adding a new devices no longer requires a restart of MAD
- Fix for #599 (idle devices not showing on status page)
- Routecalcs
  - Ad-Hoc recalculation does not remove the old calculation until finished
  - Ad-Hoc recalculation is only available when running in start.py  (short-term goal is to deprecate configmode.py once the mapping.json requirement is removed #617 and use a flag to disable workers from connecting)
  - Route recalculation now shows the progress when performing optimized recalcs. The update percentages is tiered based off the number of elements.
  - Recalculation can now occur if the data-manager is in an idle state
- GeoFences
  - Geofences drawn on MADmin map can now include spaces
  - Geofences are validated prior to saving.  This will prevent an invalid GeoFence from crashing MAD 
- Corrupted images no longer prevent DeviceControl from loading.  The corrupted image is removed and DeviceControl displays no screenshot
- Corrupted stats files no longer prevent Unhandled Exceptions with MAD.  The corrupted file is removed and a new one will be generated
- Devices rebooted through MADmin now gracefully disconnect
image

### DIFF
--- a/madmin/api/__init__.py
+++ b/madmin/api/__init__.py
@@ -28,13 +28,13 @@ class APIHandler(object):
         _logger: logger (loguru.logger): MADmin debug logger
         _modules (dict): Dictionary of APIHandlers for referring to the other API sections
     """
-    def __init__(self, logger, app, data_manager, mapping_manager, ws_server):
+    def __init__(self, logger, app, data_manager, mapping_manager, ws_server, config_mode):
         self._logger = logger
         self._app = app
         self._modules = {}
         self._app.route(BASE_URI, methods=['GET'])(self.process_request)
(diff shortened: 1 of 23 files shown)


sn0opy:
Was a MAD restarted needed at all? You only had to apply settings to make it work if I'm not mistaken (hence the warning when an "unknown" connects)

Expl0dingBanana:
MAD still had to be restarted due to self.__mapping not being updated in the routemanager.

Expl0dingBanana:
You can now save Geofences created in the UI that have a space in the name

PR #620, opened 2019-12-17
add 0.163.0 32bit addresses

nepixl:


sn0opy:
0.161.1 is the current forced version. Please don't remove supported versions.

PR #626, opened 2019-12-19
Multi thread job processing

cecpk:
Start more than one job the same time.

Use config.ini argument job_thread_count to set the thread amount (Default:2)


clburlison:
thank you. thank you. thank you. Just updated pogodroid on 30+ devices using 4 threads without a hitch. hopefully the next pogo update doesn't take an hour to apply 😉

Issue #566, opened 2019-11-26
Make MAD non-blocking for other areas on startup, if one of them needs route calculation


sn0opy:
Just to make sure we don't forget this in the future.

When MAD is restarted with at least one area calculating a new route, the whole starting process is delayed until calculation has finished.

DrMurx:
I’ve a proposal on that one which I’d like to implement as soon as #539 is in master. 

In a nutshell, the route calculation would use an iterative approach, starting with a quick but not very optimal route and refining it continuously  until a satisfactory area size to path length ratio has been reached, or whatever other criterion should be used. This could even happen in background in a low prio thread without stealing precious CPU cycles from other processing.

Doing so will reduce delay if the route is needed immediately after startup or area creation, at the price of the first round being a little less effective.

Expl0dingBanana:
A basic version of this is implemented into master.  As soon as the route is re-calculated (if optimized is required) it will disconnect / reconnect and use the new route.  If you wish to modify it to complete the route a new PR will need to be created.

PR #639, opened 2019-12-25
Route calculation on startup is non-blocking

Expl0dingBanana:
Route calculation is almost non-blocking on boot.  If a route is not calculated it will perform a quick calculation.  Once the quick recalculation is finished a new thread is started to perform ad-hoc recalculation which is non-blocking.

Implementation for #566


Expl0dingBanana:
image

Issue #618, opened 2019-12-16
Make route filed editable/deleteable


ChaosEmpire:
Before the Database migration it was easy to delete a route calc file, as you knew the name, now you need to find the corresponding id from 3 tables, and at the end you dont know if somethign will break mad when you do it directly in sql cause of dependencies

also important, there is no easy way to edit the calec route, like we have been doing before database migration

The way geofences can be opened in madmin and edited would be perfect for routefiles
that way we would have an user friendly way of deleting routes..which else stay in database forever, and we can copy paste in route, to change starting area or optimize in other ways the route

Expl0dingBanana:
Would getting the route calc by performing GET to /api/area/routecalc to get the list and POST to /api/area/routecalc be sufficient?

ChaosEmpire:
From my point of view, its not sufficent

A very easy way to remove old routes
and or edit route entries is missing in the GUI

it would be the sufficient replacement for txt editing before db migration

API calls are for advantage users, or automatisation from my point of view
A regular user should be able to do basic tasks via the gui

geofence website is a very good example how i would expect a route webpage

PR #643, opened 2019-12-29
Routecalc in MADmin

Expl0dingBanana:
Implementation for #618 .  Routecalcs are now editable in MADmin.  They are accessible on the area page on the route icon

image

image

### DIFF
--- a/madmin/routes/config.py
+++ b/madmin/routes/config.py
@@ -45,6 +45,7 @@ class config(object):
             ("/settings/ivlists", self.settings_ivlists),
             ("/settings/monsearch", self.monsearch),
             ("/settings/shared", self.settings_pools),
+            ("/settings/routecalc", self.settings_routecalc),
             ("/settings/walker", self.settings_walkers),
             ("/settings/walker/areaeditor", self.settings_walker_area),
             ("/reload", self.reload)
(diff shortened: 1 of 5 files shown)


ChaosEmpire:
While searching for the Tab ROUTES i detected the route button in AREAS
edit button works as expected, edited 2 routes successful

Missing:
As there is no route tab , like for geofence, you cant delete routes, or create one by just adding them

FimoJones:
Looks good. It is a very nice improvement and from my perspective, works as intended.

Expl0dingBanana:
image

Expl0dingBanana:
@ChaosEmpire RouteCalcs do not have a name and take their name from the area.  For that reason we do not have a routecalc page.  If you wish to manually edit a routecalc you will need to click on the route icon.  If you wish to assign a different routecalc to the area you will need to use the API.

ChaosEmpire:
@expl0dingbanana

without an own website, you cant delete routes

before database migration, you have been able toi change routes, based on file NAMES

for example, i saved a good working route, when i tested things with overlapping areas and geofence exclude
or for cday things

so i have build several routes, which i was able to change, just by changing the name in the area window

this feature got lost during db migration

so i no longer can choose a route to use in areas, 
and i also cant delete oldroutes, 

its wrong to say routes have had no names, as the name was, like the geofence, the filename you used for that route
like exploding.calc
banana,calc
explodingbanana.calc

3routes you have been able to switch between

for cdays i just took a well calculated route, removed some parts, and have had good routes for some cday specific areas
or i optimized a route myself, and have had the original to compare and also test by switching the routes

its no longer possible at the moment

sn0opy:
It's still possible. Just modify them in the DB instead.

Expl0dingBanana:
Creating a new routecalc
create a file called routefile.txt 
  (https://gist.github.com/Expl0dingBanana/4182a941af7f04b46a7991fe39ffd24e)
curl -x POST -d @routefile.txt http://localhost:5000/api/routecalc
  Check the response for X-URI (you will use this in the next step)

Set routecalc to area
curl -x PATCH -d '{"routecalc": <X-URI>}' http://localhost:5000/api/area/<area_id>
OR
login to DB and set routecalc (use the correct table based on the area) to X-URI

PR #652, opened 2020-01-03
move internal version from version.json to DB

### LINKED ISSUE #443: version.json causing issues when running multiple instances out of one directory
When running multiple, separate instances of MAD out of the same directory, utilizing the `--config` and `--mappings` options, the current approach of tracking major changes through a single entry in version.json is not suitable.

For every instance's first run after an update which increases the version, you have to manually set back the version number in version.json so all the instances can go through the update procedures.

I haven't had the time to look into solutions. As these version steps are usually related to changes in the database or config files, maybe it would be suitable to store the version number in database? That would make sure that database changes are run once per database.

### LINKED ISSUE #625: version.json location
Is it possible to get version.json moved into configs or files?

It's possible this only affects me with the way I have my docker container setup but I don't have /usr/src/app as a volume so it doesn't persist between image updates which is causing problems (the instance column in trs_status is being removed, not sure what else might be going wrong!).

Cheers

crhbetz:
resolves #625 (also resolves #443 😃 ) and should enable multi-instance, multi-DB from a single MAD directory.

As we've pretty much moved everything possible into the database now, it's just logical to make this change.

Tested locally by playing around with the `current_version` value in `utils/version.py` and looking at the log outputs and database and subsequently moved to my production instances without any apparent issues.

Please notify or edit especially if there's any issue with the changes to `db/DbWrapper.py` - it seems to work but I'm not sure if it's correctly implemented within the system of `db/` files and classes/methods.

### DIFF
--- a/db/DbWrapper.py
+++ b/db/DbWrapper.py
@@ -1028,3 +1028,21 @@ class DbWrapper:
             }
             res = self._db_exec.autoexec_insert('madmin_instance', instance_data)
             return res
+
+    def set_mad_version(self, version):
+        query = "INSERT INTO versions (versions.key, val) VALUES('mad_version', {})".format(int(version))
+        res = self.execute(query, commit=True)
+        return res
+
+    def get_mad_version(self):
+        query = "SELECT val FROM versions where versions.key = 'mad_version'"
+        res = self.execute(query)
+        if res:
+            return int(res[0][0])
+        else:
+            return None
+
+    def update_mad_version(self, version):
+        query = "UPDATE versions SET val = '{}' WHERE versions.key = 'mad_version'".format(int(version))
+        res = self.execute(query, commit=True)
+        return res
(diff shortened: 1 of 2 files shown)


Expl0dingBanana:
We might want to break this into a MAD version and DB version.  If there is any modification of data we want to be able to do that per-installation and the DB changes would be for all instances.

crhbetz:
I can't quite follow, sorry. What do you mean by "modification of data"? Are you referring to data that's not in DB yet and thus could be present in multiple instances?

If so, how could different installations be **permanently** identified with `status_name` config option not being required? I don't know what would be practical here - store their paths in DB? Tell me what to do and I'll try to figure it out .. 😅

Expl0dingBanana:
The more I think about it just one version is fine.  We just need to make sure when we update version.py it can account for multi-instance systems.

PR #653, opened 2020-01-03
Rotuecalc / GeoFence validation improvements

Expl0dingBanana:


Expl0dingBanana:
Added a check for empty lines.  They shouldnt exist in the routecalc but they shouldnt cause errors with MAD startup.  Extra logging has been added around these resources to help diagnose any future issues.

PR #655, opened 2020-01-03
Check mainscreen before clearing box

stonedDiscord:
Sometimes the worker thinks it's clearing items while really, it is just turning the map because it closed the team rocket dialog instead of opening the communicator menu.

clear_box assumes that it's on the mainscreen when it starts and this PR adds a check for that.

Haven't had it get stuck since making this change.


cecpk:
fixed with pr #662

PR #658, opened 2020-01-04
Added raid costume support to db and webhook

JimBim89:
Added costume support for webhook type "raid" which is supported by other receivers.
Also modified files to create db column and pick up information to fill the needed data.

This additional costume information for raids is useful for the currently costumed raids.


chr2002qw:
Here could be a small bug. If a costumed mon appears on a gym, the costume ID gets saved to table. But, if later a non costumed appears, the costume ID is not set to NULL. So Rocketmad trieds to generate a costumed mon. But it doesnt extist in assetes and this placeholder icon is shown on gym. Saw this on Mon 303 as example.

PR #659, opened 2020-01-05
Initial development of MADmin hosted APKs

Expl0dingBanana:
This allows MAD to host any uploaded APKs.  This allows for local file-transfers when installing APKs for ATVs.

Testing:
Upload a file (there are no version checks or validation you are using the correct type [pogo, pd, rgc])
wait for the screen to refresh.  Once refreshed you should see your file.  You can download the file @
`/api/mad_apk/<apk_type>/<arch>/download` (ie `/api/mad_apk/pogo/armeabi-v7a/download`).  Authorization for this page are the same requirements as RGC / PD.
To-Do list (future releases)
Move the upload to the API
Update Jobs to use these local files
Have MAD auto-find newer versions and import (probably a config flag)

image

### DIFF
--- a/db/DbSanityCheck.py
+++ b/db/DbSanityCheck.py
@@ -7,7 +7,14 @@ class DbSanityCheck:
 
     def __init__(self, db_exec: PooledQueryExecutor):
         self._db_exec: PooledQueryExecutor = db_exec
+        self.failing_issues = False
+        self.supports_apks = False
 
+    def check_all(self):
+        self.ensure_correct_sql_mode()
+        self.validate_max_allowed_packet()
+        if self.failing_issues:
+            sys.exit(1)
 
     def ensure_correct_sql_mode(self):
         """
@@ -25,4 +32,15 @@ class DbSanityCheck:
             logger.error("Please drop those settings: {}.", ", ".join(detected_wrong_modes))
             logger.error(
                 "More info: https://mad-docs.readthedocs.io/en/latest/common-issues/faq/#sql-mode-error-mysql-strict-mode-mysql-mode")
-            sys.exit(1)
+            self.failing_issues = True
+
+    def validate_max_allowed_packet(self):
+        query = "SELECT @@global.max_allowed_packet"
+        res = self._db_exec.autofetch_value(query)
+        # Not sure what a good number is.  Based on the current size 150MB seems reasonable for APK size
+        if res < 1024*1024*150:
+            logger.error("max_allowed_packet may need to be adjusted to use the MAD APK feature")
+            logger.error("MAD will function without this being set but you will be unable to upload and server MAD APK files")
+            logger.error("More info can be found @ https://dev.mysql.com/doc/refman/8.0/en/program-variables.html")
+        else:
+            self.supports_apks = True
--- a/db/DbSanityCheck.py
+++ b/db/DbSanityCheck.py
@@ -1,6 +1,7 @@
 import sys
 from utils.logging import logger
 from db.PooledQueryExecutor import PooledQueryExecutor
+from utils import global_variables
 
 class DbSanityCheck:
     blacklisted_modes = "NO_ZERO_DATE NO_ZERO_IN_DATE ONLY_FULL_GROUP_BY"
@@ -37,10 +38,9 @@ class DbSanityCheck:
     def validate_max_allowed_packet(self):
         query = "SELECT @@global.max_allowed_packet"
         res = self._db_exec.autofetch_value(query)
-        # Not sure what a good number is.  Based on the current size 150MB seems reasonable for APK size
-        if res < 1024*1024*150:
+        if res < global_variables.CHUNK_MAX_SIZE:
             logger.error("max_allowed_packet may need to be adjusted to use the MAD APK feature")
-            logger.error("MAD will function without this being set but you will be unable to upload and server MAD APK files")
+            logger.error("MAD will function without this being set but you will be unable to upload and serve MAD APK files")
             logger.error("More info can be found @ https://dev.mysql.com/doc/refman/8.0/en/program-variables.html")
         else:
             self.supports_apks = True
--- a/madmin/routes/apks.py
+++ b/madmin/routes/apks.py
@@ -1,13 +1,11 @@
-from flask import (send_from_directory, render_template, request, jsonify, redirect, url_for, flash)
+from flask import (send_from_directory, render_template, request, jsonify, redirect, url_for, flash, Response)
 from madmin.functions import (auth_required, get_quest_areas)
 from utils.functions import (generate_path)
 from utils.MappingManager import MappingManager
 from utils.logging import logger
-import utils.apk_util
-from werkzeug.utils import secure_filename
+from utils import apk_util
 from utils import global_variables
-import apkutils
-import os
+import json
 
 class apk_manager(object):
     def __init__(self, db, args, app, mapping_manager: MappingManager, deviceUpdater):
@@ -22,6 +20,7 @@ class apk_manager(object):
         routes = [
             ("/apk", self.mad_apks),
             ("/apk_upload", self.upload_file, ['POST']),
+            ("/apk_update_status", self.apk_update_status),
         ]
         for route_def in routes:
             if len(route_def) == 2:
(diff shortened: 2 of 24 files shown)


madmin/routes/apks.py
```diff
@@ -0,0 +1,105 @@
...
+                        if filestore_id:
+                            delete_data = {
+                                'id': filestore_id
+                            }
+                            self._db.autoexec_delete('filestore', delete_data)
+                        insert_data = {
+                            'filename': filename,
+                            'data': apk_upload.read()
+                        }
+                        new_id = self._db.autoexec_insert('filestore', insert_data)
```
fosJoddie: What's the rationale for apk files being saved in the DB instead of being saved to disk directly?

On downside of writing large files to DB is that it will cause more issues with people asking for help on chaning their sql server settings.
additionally, have you done any testing for memory usage when serving files over DB?

db/DbSanityCheck.py
```diff
@@ -25,4 +32,15 @@ def ensure_correct_sql_mode(self):
...
-            sys.exit(1)
+            self.failing_issues = True
+
+    def validate_max_allowed_packet(self):
+        query = "SELECT @@global.max_allowed_packet"
+        res = self._db_exec.autofetch_value(query)
+        # Not sure what a good number is.  Based on the current size 150MB seems reasonable for APK size
+        if res < 1024*1024*150:
+            logger.error("max_allowed_packet may need to be adjusted to use the MAD APK feature")
+            logger.error("MAD will function without this being set but you will be unable to upload and server MAD APK files")
```
fosJoddie: Typo. Change to
> logger.error("MAD will function without this being set, but you will be unable to upload and serve APK files")

fosJoddie:
An issue with this is that you have to expose madmin to the internet if you want to support off-site deployed atvs. - we already have a way for the RGC-controller end-point to push apks across without exposing the madmin webpage. - we gotta be careful to still allow users to carry on using this method even for ATVs

madmin/routes/apks.py
Expl0dingBanana: The reason was to support multi-instance deployments without having to upload the file to each instance.  In terms of memory it probably needs to be switched to a stream so it uses less.

db/DbSanityCheck.py
Expl0dingBanana: This has been corrected.

fosJoddie:
That sounds good, just wanted to ne sure it was thought of 💪🥳

Expl0dingBanana:
A new type of job has been added, Smart Update.  This feature is used in conjunction with the new APK page as they must be present in MADmin to work.  This job will validate the architecture and version on the device.  If MAD has a new APK that matches it will install it on the device.

Expl0dingBanana:
APK Successfully updating
image

APK does not need to be updated
image

MAD does not contain APK (not in MAD or incorrect architecture)
image

clburlison:
Just used this branch and the new Smart update to deploy the latest PogoDroid. Checking to see if the update is required is a nice touch 👌 
IMO the "MADmin APKs" page should be under the Jobs tab or Settings > new tab for "MADmin APKs"

clburlison:
General feedback
* When uploading a 200 MB zip file (I wanted to see what would happen) the connection crashes ungracefully.
https://gist.github.com/clburlison/3fffd796f9ed44b396a6f04484e9da48
* When clicking the MADmin APKs link from the menu bar it takes us to the apk status page but also gives us an upload window.
* I managed to upload the `arm64-v8a` architecture apk in the place of the `armeabi-v7a` arch. The following looks like a possible solution on how we could validate the apk arch after upload. I tested this tested with Pogo to determine 32/64 and it works for PogoDroid, however the logic doesn't hold for RGC. It _should_ be stable enough to do the apk arch validation check but :noidea: meme.
 
```python
>>> import apkutils
>>> temp = apkutils.APK("./Downloads/pogo64.apk")
>>> any("armeabi-v7a" in d.get('name') for d in temp.get_files())
False
>>> any("arm64-v8a" in d.get('name') for d in temp.get_files())
True

>>> temp = apkutils.APK("./Downloads/PogoDroid.apk")
>>> any("armeabi-v7a" in d.get('name') for d in temp.get_files())
True
>>> any("arm64-v8a" in d.get('name') for d in temp.get_files())
True
>>> 

>>> temp = apkutils.APK("./Downloads/RemoteGpsController.apk")
>>> any("armeabi-v7a" in d.get('name') for d in temp.get_files())
False
>>> any("arm64-v8a" in d.get('name') for d in temp.get_files())
False
```

Expl0dingBanana:
image
This image shows the MADmin search and download functionality.  There are plans to add a 'Update everything' button so it will search and download all with a single click.

Expl0dingBanana:
@clburlison 
Issues 1 and 2 have been fixed.  For the third its inconsistent so I think it will be up to the user to make sure they are using the correct values.  I am hoping people will opt to use the auto-downloader and not mess with manually uploading APKs :)

DrMurx:
@Expl0dingBanana Please do NOT store binary data in database blobs. This evicts precious cached data from other tables. There's nothing wrong in storing files in the filesystem ;-) 

We already have an uploads folder, and I'd suggest to store the data there - maybe using filenames like `pogo_$file_id.fetched.apk`, so you can simply exclude those files from the jobs list so that people don't get confused.

wraythex:
*On 2020/01/08 at 3 UTC:
I just tested PR659 on my system.  The wizard hat worked normally.  Since I already updated PD manually, I could only test pushing RGC.
I first ran a "readout rgc version" job on 5 devices (1,2,6,7,8 = atv101,atv102,atv106,atv107,atv108)
It returned version 1.9.15 on 1,2,6,8.  For some reason 7 failed to return a version.
I then ran the "smart update - rgc" to 1,2,6,7,8.  It succeeded on 1,2,6,7 and failed on 8.  (error: "Unable to determine the architecture of the device")
I then ran the "readout rgc version" job again, and 1,2,6 reported 1.9.18/1.9.15.  The job failed on 7 and 8 with the following:
```[01-07 21:29:19.76] [    apk_updater-1] [            updater:544 ] [   ERROR] Error while getting response from device - Reason: 'NoneType' object has no attribute 'replace'
[01-07 21:29:19.76] [    apk_updater-1] [            updater:544 ] [   ERROR] Error while getting response from device - Reason: 'NoneType' object has no attribute 'replace'
[01-07 21:29:19.76] [    apk_updater-1] [            updater:303 ] [   ERROR] Job jobType.PASSTHROUGH could not be executed successfully - Device atv107 - File/Job dumpsys package de.grennith.rgc.remotegpscontroller | grep versionName (ID: 1578454055)
[01-07 21:29:19.76] [    apk_updater-1] [            updater:303 ] [   ERROR] Job jobType.PASSTHROUGH could not be executed successfully - Device atv107 - File/Job dumpsys package de.grennith.rgc.remotegpscontroller | grep versionName (ID: 1578454055)```
(sometimes this above message appears a few times, and then a later retry of the job will succeed.  So, not sure why this happens or why it sometimes eventually works)
Manually checking the version through adb shell, I see that 7 has RGC version 1.9.18/1.9.15, but atv108 still only has 1.9.15 installed.
I was able to manually update RGC on atv108 through adb shell.
All devices either stayed connected or automatically reconnected on their own.  They all are doing work like normal.

Expl0dingBanana:
@DrMurx The DB was chosen to support multiple instances without having to upload the file for each server.

DrMurx:
> @DrMurx The DB was chosen to support multiple instances without having to upload the file for each server.

@Expl0dingBanana Multiple instances can also simply refer to the very same upload directory. The disadvantages of storing big binary blobs in a database are really outweighing any advantages of a simplified multi-instance setup. Please don't do it.

domizei385:
@DrMurx I tried to search, how MariaDB operates when dealing with binary data. Do you have any references to what you describe?

JabLuszko:
Also please don't store files for sake of backup. It's hard to convince people to do backups, it will be even harder if they have to exclude some tables and/or backups will take 1.2 GB and 6 hours to transfer.

DrMurx:
> @DrMurx I tried to search, how MariaDB operates when dealing with binary data. Do you have any references to what you describe?

@domizei385 It's by no means different from storing the very same amount of text. 

We're talking about chunks of 8MB which are written as a single transaction. Given that most people probably haven't tuned there MySQL settings and have `innodb_buffer_pool_size` at the default 128 MB, this will evict at least 1/16 of the buffered data and index pages, most likely more. 

Depending on how many devices you operate, your `pokemon` table is undergoing constant alteration while at the same time being regularly polled by the webhook thread. Its performance depends on keeping as much data as possible in memory. Any eviction could lead to higher IO afterwards, until the cache has recovered.

Also, keep in mind that some people operate MAD on a Raspberry PI or other low-cost devices.

DrMurx:
> Also please don't store files for sake of backup. It's hard to convince people to do backups, it will be even harder if they have to exclude some tables and/or backups will take 1.2 GB and 6 hours to transfer.

Dude, you just opened my eyes why my MAD crashes every day when the backup starts... 🙄

crhbetz:
Regarding multiple instances: With most things, including version.json, now residing in database, I don't think there's a lot of reasons left to have multiple MAD folders at all. The only thing I can think of is if you have multiple devices with the same origin name on different instances, causing conflicts with the .position and .stats files.

So imo storing files is okay; if it takes too much space for multi-MAD-folder-setups, one could point the affected user to just using one folder instead.

muckelba:
For me personally, i use different MAD DBs/Dirs because of the lack of proper geofence filters for webhooks. Everything else (like frontend) is able to filter based on geofences and could happen in the same DB

sn0opy:
Due to some bigger changes to the whole code base, we're going to merged this into MAD already. Feel free to open up issues if you encounter any issues.

PR #660, opened 2020-01-06
Refactor screen handling

Grennith:
Basically bugfixing for multiple issues

- Moved ScreenType enum to ocr/screen_type.py and also commented the values (which I hope I got right...)
- Changed naming of WorkerBase._check_windows to WorkerBase._ensure_pogo_topmost along with most of its body. Removed the == ScreenType.POGO check in the while-condition in order to handle all screens. This is achieved by also lifting conditions for the same-screen-checks to trigger for all ScreenType != ERROR
- Refactored WordToScreenMatching.detect_screentype with 2 new methods splitting up the previous logic (namely __evaluate_topmost_app and __handle_screentype). While doing so, I added all possible enum values to __handle_screentype in order to at least log the types. This construct should be thought over anyway considering WorkerBase._ensure_pogo_topmost handles certain enum values (e.g. .SN) while WordToScreenMatching.__handle_screentype handles stuff like PTC login... it's spaghetti
- Fixed an issue in WebsocketServer.__register where reconnects were handled with an asyncio.sleep... holding a lock
- Removed redundant close() calls on the websocket client in WebsocketServer.__register
- Added typing to WebsocketServer.__register. There was one case where a tuple was returned and another where nothing was returned at all. Meanwhile, the callee (WebsocketServer.handler) relied on a True/False decision on whether to continue work or stop handling the incoming connection
- Added logging regarding WebSocketClientProtocol.close() by wrapping it in a method
- Added usage of the WebSocketServerProtocol's close_timeout parameter - although that should be initialised with a value of 10 according to https://websockets.readthedocs.io/en/stable/_modules/websockets/server.html
- Added logging of exceptions regarding tesseract usage
- Added counting for both restart_pogo usages in WorkerBase in order to circumvent stackoverflows by calling reboot at a count of 2 restarts that way (sorta dirty workaround)

### DIFF
--- a/ocr/pogoWindows.py
+++ b/ocr/pogoWindows.py
@@ -874,9 +874,15 @@ class PogoWindows:
             return None
 
         for text in texts:
-            globaldict = self.__internal_get_screen_text(text, identifier)
+            try:
+                globaldict = pytesseract.image_to_data(text, output_type=Output.DICT, timeout=40,
+                                                       config='--dpi 70')
+            except Exception as e:
+                logger.error("Tesseract Error for device {}: {}. Exception: {}".format(str(identifier),
+                                                                                       str(globaldict), e))
+                globaldict = None
             logger.debug("Screentext: {}".format(str(globaldict)))
-            if 'text' not in globaldict:
+            if  globaldict is None or'text' not in globaldict:
                 continue
             n_boxes = len(globaldict['level'])
             for i in range(n_boxes):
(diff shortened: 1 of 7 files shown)


ocr/pogoWindows.py
```diff
@@ -667,7 +671,11 @@ def __internal_get_inventory_text(self, filename, identifier, x1, x2, y1, y2):
         gray = cv2.resize(gray, dim, interpolation=cv2.INTER_AREA)
         cv2.imwrite(temp_path_item, gray)
         with Image.open(temp_path_item) as im:
```
fosJoddie: corrupt/partial file transfers might cause Image.open to fail here, I suggest moving the try/catch block outside this `with` block

fosJoddie: actually the file reference I meant is probably read on line 659 (first line of method)

PR #665, opened 2020-01-07
Add s2 cells for raid route coordinates

sn0opy:
image

Baiscally what this does is, it adds surrounding L15 S2cells to each raid route coordinate to make it more clear what is being scanned and what not. 

Todo but not actually blocking this to be merged:
- render cells in JS, not in Python on the server
- use correct, dynamic radius depending on the latitude (discussed on Discord)


wraythex:
Tested this PR on my setup and it worked well.

PR #683, opened 2020-01-14
Improved amount of device screens for higher resolutions

JimBim89:
Without these modifications you could only have a maximum of 3 screens,
no matter how much white-space you had or which zoom level you used.

Now its responsive to zoom and adds more than 3 screens on higher resolutions.


sn0opy:
Nice addition, thx

Issue #646, opened 2019-12-31
Pokestop Coloring in MADmin


marcb1387:
can we make Pokestops change colors based on if they have gotten a field research task for the day.

Red stop color - no field research
Blue stop color - it's has field research

it would make seeing which stops did not get field research easier to visualize.

PR #687, opened 2020-01-15
pokestops without quests now appear red on the map

Expl0dingBanana:
Manually updated the database to show a single pokestop with a quest
image

### DIFF
--- a/db/DbWrapper.py
+++ b/db/DbWrapper.py
@@ -19,6 +19,8 @@ from db.DbStatsSubmit import DbStatsSubmit
 from db.DbStatsReader import DbStatsReader
 from db.DbWebhookReader import DbWebhookReader
 
+import re
+
 
 class DbWrapper:
     def __init__(self, db_exec, args):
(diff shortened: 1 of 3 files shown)


Expl0dingBanana:
Implementation for #646

madmin/static/js/madmin.js
```diff
@@ -757,16 +757,21 @@ new Vue({
 
       axios.get("get_stops" + urlFilter).then(function (res) {
         res.data.forEach(function(stop) {
+          console.log(stop);
```
fosJoddie: orphaned console.log, please remove/adopt

madmin/static/js/madmin.js
```diff
@@ -757,16 +757,21 @@ new Vue({
...
           if ($this.stops[stop["id"]]) {
             return;
           }
+          var color = "blue";
```
fosJoddie: assign only once: `var color = stop["has_quest"] ? "blue" : "red";`

db/DbWrapper.py
```diff
@@ -628,52 +628,53 @@ def get_mons_in_rectangle(self, neLat, neLon, swLat, swLon, oNeLat=None, oNeLon=
...
-                "active_fort_modifier": active_fort_modifier,
-                "last_updated": int(last_updated.replace(tzinfo=timezone.utc).timestamp()) if last_updated is not None else None,
-                "name": name,
-                "image": image,
-                "incident_start": int(incident_start.replace(tzinfo=timezone.utc).timestamp()) if incident_start is not None else None,
-                "incident_expiration": int(incident_expiration.replace(tzinfo=timezone.utc).timestamp()) if incident_expiration is not None else None,
-                "incident_grunt_type": incident_grunt_type
+        pokestops = self.autofetch_all(query + query_where, args=tuple(args))
+
+        for pokestop in pokestops:
```
fosJoddie: super out of scope for this PR, but since you've already refactored parts of the select code anyhow:
Are we using the returned dict from this method anywhere? 
Cause for the map we do the select, convert it to a dict, then we iterate that dict to make a list of coords again.

PR #734, opened 2020-02-05
add 0.165.1 32bit

nepixl:


nepixl:
APK from APKPure:
https://apkpure.com/de/pokemon-go/com.nianticlabs.pokemongo/variant/0.165.1-APK

PR #738, opened 2020-02-07
Enable/Fix time sorting on status page and add optional checkmark

JabLuszko:
Columns with timestamps are now correctly sortable.

Moved ajax call to DataTable functions rather than some extra stuff we got there.

There is optional ✔️in Last Update column, set at the bottom of page, stored in Local Storage (so per browser) - this is just for people who just want to take a quick glance at status and see if it's working.
Fully optional, so if you don't put anything there it won't ever show up.
image

### DIFF
--- a/static/madmin/templates/status.html
+++ b/static/madmin/templates/status.html
@@ -12,13 +12,11 @@
 </style>
 <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.22.2/moment.min.js"></script>
 <script>
-    var dataTable;
-    function setGrid(tableGridHtmlId, gridData) {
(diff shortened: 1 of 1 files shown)


muckelba:
Working good. Thx! 👍

cecpk:
@JabLuszko did you test that changes with new darkmode too?

JabLuszko:
@cecpk nope, darkmode is newer than this :D. 
It's a green checkmark and one input box without any special classess, so it should inherit w/e is there aka "should work".

Axiangel:
Finally came around to test this. Works great here. 
Not a second to soon to get proper sorting. Thanks JabLuszko! <3

JabLuszko:
@cecpk works with darkmode.
image

PR #740, opened 2020-02-10
Add support for quest condition using specific items to evolve

fosJoddie:
Includes French and German translations


locale/de/LC_MESSAGES/quest.po
```diff
@@ -18,329 +18,338 @@ msgstr ""
...
-#: questGen.py:195
+#: questGen.py:211
 msgid "Use an item to evolve a Pokemon"
 msgstr "Verwende ein Item um ein Pokemon zu entwickeln"
 
-#: questGen.py:197
+#: questGen.py:216
+#, python-brace-format
+msgid "Use a {item} to evolve a Pokemon"
+msgstr "Benutze einem {item}, um ein Pokemon zu entwickeln"
```
JimBim89: Instead of "Benutze einem {item}, um ein Pokemon zu entwickeln",
you should use "Benutze {item}, um ein Pokemon zu entwickeln".

fosJoddie: it's translated to that because that's how it was written in-game - are you sure the sentence works better without `einem`?  (which I now I should be `einen`

image

muckelba: If it's `stein`, use `einen`.

JimBim89: Yeah, because of the different evolution items which could be used here. Not all would read correctly, some would be wrong. So leaving it out should work for all of them without loosing valuable information.

JimBim89:
Also, "Battle 3 Team Rocket Grunts" and "Earn 5 hearts with your Buddy" are currently not translated into German at all. Don't know if we should integrate them here or make a separate issue/pr.

fosJoddie:
> Also, "Battle 3 Team Rocket Grunts" and "Earn 5 hearts with your Buddy" are currently not translated into German at all. Don't know if we should integrate them here or make a separate issue/pr.

They aren't translated because nobody provided a German translated version of  `types.json` - I don't speak German so I am not up for this task, but I'm happy to integrate the missing file from anyone who has the time - it's only ~35 tasks/sentences

locale/de/LC_MESSAGES/quest.po
fosJoddie: I see, I will update and remove `ein`

PR #741, opened 2020-02-10
Quest fixes

### LINKED ISSUE #746: Quest Worker not clearing inventory
Sometimes (seen it a few times now) the following happens:

This worker tries to spin a stop, can't, because of inventory is full, and will never suceed to empty its inventory.

https://pastebin.com/d6qSvWs5

as far as i can tell by experimenting, when it arrives in this state, you can't assign it to a new walker anymore, as it's stuck in this "unprocessed event" + "can't do that" loop.

(mad restart helps)

### LINKED ISSUE #749: Exception during internal cleanup
I'm getting the following Exception during Questscan when MAD is trying to cleanup a device:

https://pastebin.com/J18XBrzb

crhbetz:
Update 2020-02-17:

Completely reworked and tested this. Should now handle the following, as described in the commit notes:

* fix #746: restart PoGo if clear tasks constantly fail
* fix #749: handle all WebsocketWorker* exceptions
* TODO: less magic numbers
* a little more logging for more info
* detect full quest log (FortSearchResultTypes.FULL)
* avoid softban deadlock by closing the stop after every spin attempt
* fix stops being silently skipped, handle timedelta issue better
* add GMO / 106 handling to _wait_data_worker (needed for spinnable check)

The goal is to increase efficiency and consistency of the quest scan mode by fixing a couple of issues and improving the overall working logic here and there.

Over a weekend of testing, I achieved more consistent scanning which translates into more quests being scanned in a given time frame.

### DIFF
--- a/mapadroid/worker/WorkerQuests.py
+++ b/mapadroid/worker/WorkerQuests.py
@@ -19,7 +19,8 @@ from mapadroid.utils.logging import logger
 from mapadroid.utils.madGlobals import (
     InternalStopWorkerException,
     WebsocketWorkerRemovedException,
-    WebsocketWorkerTimeoutException
+    WebsocketWorkerTimeoutException,
+    WebsocketWorkerConnectionClosedException
 )
 from mapadroid.websocket.AbstractCommunicator import AbstractCommunicator
 from mapadroid.worker.MITMBase import MITMBase, LatestReceivedType
@@ -36,6 +37,13 @@ class FortSearchResultTypes(Enum):
     LIMIT = 5
     UNAVAILABLE = 6
     OUT_OF_RANGE = 7
+    FULL = 8
+
+
+class ClearThreadTasks(Enum):
+    IDLE = 0
+    BOX = 1
+    QUEST = 2
 
 
 class WorkerQuests(MITMBase):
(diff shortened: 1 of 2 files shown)


crhbetz:
Little update:
After evaluation, I'm currently completely re-doing this already seeing quite satisfactory results. I'll update the PR _after_ I'm done with my own testing :)

crhbetz:
Reworked and updated, see first comment.

fosJoddie:
@crhbetz could you kindly rebase this against master?

crhbetz:
@fosJoddie done :-)

crhbetz:
Got some good, some neutral, no negative feedback over the last couple of days, and running it myself for over a week now.
As we're probably not gonna get any more testers, I'd personally consider this ready for merge. 🤷‍♂

fosJoddie:
Ran fine for me in the last 24 hours, going to give it another night for comparison - the current pogo version likes to go blank map a lot when teleporting, and it seems from one night of testing that this PR does a better job of discovering that this is happening.

mapadroid/worker/WorkerQuests.py
```diff
@@ -619,6 +650,9 @@ def _current_position_has_spinnable_stop(self, timestamp: float):
                 fort_type: int = fort.get("type", 0)
                 if fort_type == 0:
                     self._db_wrapper.delete_stop(latitude, longitude)
+                    # TODO: if you know what fort_type 0 means, please edit
+                    # this logline accordingly :)
+                    logger.warning("Can't spin the stop because of fort_type 0")
```
fosJoddie: fort type 0 are gyms - this is why the if line deletes the stop from the DB. This shouldn't actually happen anymore though, since PD blocks gyms opening up when they are clicked. Code should be left in here for those special rare cases tho.

TLDR; change text to "Tried to open a stop but found a gym instead"

PR #759, opened 2020-02-25
Cause a timeout if data received is too far away from intended scan location

crhbetz:
... Thus a device will eventually restart pogo if it seems to be permanently stuck at a location. There have been reports from multiple users, me among them, of this happening.
MAD can't currently handle it because pogo doesn't freeze, so PogoDroid is still able to transmit related data, but from a wrong location.
Testing shows that a pogo restart seems to be enough to fix this.

### DIFF
--- a/mapadroid/worker/WorkerBase.py
+++ b/mapadroid/worker/WorkerBase.py
@@ -28,6 +28,7 @@ from mapadroid.utils.routeutil import check_walker_value_type
 from mapadroid.websocket.AbstractCommunicator import AbstractCommunicator
 from mapadroid.worker.AbstractWorker import AbstractWorker
 from mapadroid.utils.geo import get_distance_of_two_points_in_meters
+from mapadroid.utils.s2Helper import S2Helper
 
 
 class WorkerBase(AbstractWorker):
@@ -1205,11 +1206,24 @@ class WorkerBase(AbstractWorker):
         else:
             data_to_check = "forts"
         lat_sum, lng_sum, counter = 0, 0, 0
-        for cell in data:
-            for element in cell[data_to_check]:
-                counter += 1
-                lat_sum += element["latitude"]
-                lng_sum += element["longitude"]
+
+        if data_to_check == "forts":
+            for cell in data:
+                if cell[data_to_check]:
+                    cell_id = cell["id"]
+                    if cell_id < 0:
+                        cell_id = cell_id + 2 ** 64
+                    lat, lng, alt = S2Helper.get_position_from_cell(cell_id)
+                    counter += 1
+                    lat_sum += lat
+                    lng_sum += lng
+        else:
+            for cell in data:
+                for element in cell[data_to_check]:
+                    counter += 1
+                    lat_sum += element["latitude"]
+                    lng_sum += element["longitude"]
+
         if counter == 0:
             return None
         avg_lat = lat_sum / counter
--- a/mapadroid/worker/WorkerBase.py
+++ b/mapadroid/worker/WorkerBase.py
@@ -1196,6 +1196,9 @@ class WorkerBase(AbstractWorker):
             self, self._screen_x, self._screen_y, x_offset, y_offset)
 
     def _check_data_distance(self, data):
+        max_radius = self._mapping_manager.routemanager_get_max_radius(self._routemanager_name)
+        if not max_radius:
+            return True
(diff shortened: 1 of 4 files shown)


kamieniarz:
This issue is kinda rare so I think "Data is too far away" message could be moved to info logger or maybe shortened and added to info leaving debug as it is. Great job btw ;)

crhbetz:
You wouldn't want that. The function runs in a loop until the timeout threshold is reached, thus it would be spam the logs every second or so when it happens.

kamieniarz:
So basically after few days of testing I'm not having this issue with device not moving cause mad finally treats this bahaviour as timeouts (so everything works as intended). Great job!
I think it's ready to push

mapadroid/worker/WorkerBase.py
```diff
@@ -1181,3 +1182,41 @@ def _get_screen_size(self):
...
+                lng_sum += element["longitude"]
+        if counter == 0:
+            return None
+        avg_lat = lat_sum / counter
+        avg_lng = lng_sum / counter
+        distance = get_distance_of_two_points_in_meters(float(avg_lat),
+                                                        float(avg_lng),
+                                                        float(self.current_location.lat),
+                                                        float(self.current_location.lng))
+        max_radius = self._mapping_manager.routemanager_get_max_radius(self._routemanager_name)
```
fosJoddie: routeman get max radius can return None, but this is never checked - should either return a value or check for None

mapadroid/worker/WorkerBase.py
```diff
@@ -1181,3 +1182,41 @@ def _get_screen_size(self):
         self._resocalc.get_x_y_ratio(
             self, self._screen_x, self._screen_y, x_offset, y_offset)
 
+    def _check_data_distance(self, data):
+        mode = self._mapping_manager.routemanager_get_mode(self._routemanager_name)
+        if mode in ["mon_mitm", "iv_mitm"]:
+            data_to_check = "wild_pokemon"
+        else:
+            data_to_check = "forts"
```
fosJoddie: for stops and gyms, you can legitimately get information that is further away than max radius, - has it been looked into that this edge case does not cause "too far away" timeouts? 
Example of such a situation:
image

The left circle has a row of gyms and stops that are all outside the radius, but still "in range" because of hos s2 cells work.
An alternative way to figure out this max distance is to do the calculation we do when visualizing the coverage in raid mode - and then use the distance from centre to the top left corner etc

crhbetz: Using the s2Helper stuff looks promising per se - just using the center of all `middle_of_cell`s would probably yield a more precise result, but I'm just not quite sure how to interpret the GMO data. I just looked at a couple of data samples and there always seem to be cells with nothing but their ID in the data:

`{"id":1234,"current_timestamp":9876,"forts":[],"wild_pokemon":[],"nearby_pokemon":[],"catchable_pokemon":[],"decimated_spawnpoints":[],"spawnpoints":[]}`

Would I just use those cells that contain data relevant to the scan mode?
For a real improvement, someone who knows more about those behaviours should probably look at this and tell me how to interpret the GMO data for this purpose, or just quickly rework it on their own.

cecpk:
more feedback?

mapadroid/worker/WorkerBase.py
crhbetz: I don't think there should be a case where max_radius doesn't exist - but if there were one, the whole check wouldn't make sense, thus I'd just return True in that case.

mapadroid/worker/WorkerBase.py
Grennith: The amount of empty cells is related to the distance to the player. A lot more cells are sent to the client than those populated with mons since forts, weather etc also play a role. For mon-scanning, I'd ignore those accordingly.

crhbetz: Finally got around to implement this with the latest commit. Basically just stole some code from here:
https://github.com/Map-A-Droid/MAD/blob/936ffb821ccd80d8382227904f11a3af29eb7b5a/mapadroid/db/DbPogoProtoSubmit.py#L629-L635.

I now used `get_position_from_cell` instead of `middle_of_cell` because it was used in the snippet above :D ... it seems to do exactly the same, just that the first one also returns a altitude of 0 that we don't need here?

Grennith:
Thanks :)

PR #764, opened 2020-03-05
Performance optimizations on trs_spawn & weather

DrMurx:
3 optimizations on `trs_spawn`:

* The two keys on `spawnpoint` are dropped and replaced with a single primary key.
* The column type of `spawnpoint` is aligned with `pokemon.spawnpoint_id` to increase join performance.
* Two additional keys are added to speed up some route and map related queries.

Changing a column type might affect 3rd party software, though it's not very likely. Rocketmap/Rocketmad seems to deal fine with those changes.

### DIFF
--- a/mapadroid/patcher/__init__.py
+++ b/mapadroid/patcher/__init__.py
@@ -28,6 +28,7 @@ MAD_UPDATES = OrderedDict([
     (24, 'patch_24'),
     (25, 'patch_25'),
     (26, 'patch_26'),
+    (27, 'patch_27'),
 ])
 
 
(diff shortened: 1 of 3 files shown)


DrMurx:
I'll amend the PR, the column type should be `unsigned`.

cecpk:
As talked: Patch 26 is used for Eventstuff - please update to 27. Thx

DrMurx:
@cecpk Patch is now using v27. I also modified the index to include the new event id column.

dkmur:
added changes manually as I didn't want to change version in case something else is being merged first. I see no issue with the PR so far and indeed loading time of map has improved significantly.

DrMurx:
Amended this PR with dropping a few indexes on the `weather` table.

cecpk:
call @DrMurx if any (!) problems occur after pull and starting MAD :)

PR #769, opened 2020-03-11
Prevent rare loop with ScreenType.ERROR/FAILURE

crhbetz:
Had a device endlessly loop with ScreenType.ERROR/ScreenType.FAILURE,
logging only: `[ WARNING] Something wrong with screendetection or pogo failure screen`.

It has to be assumed that `self._loginerrorcounter` somehow was incremented over 2, thus MAD was
unable to enter the appropriate routine because it only checked `self._loginerrorcounter == 2`.
I changed this to check `self._loginerrorcounter > 1` and to reset the loginerrorcounter to 0 once the routine ran to prevent this rare condition.

### DIFF
--- a/mapadroid/worker/WorkerBase.py
+++ b/mapadroid/worker/WorkerBase.py
@@ -609,7 +609,7 @@ class WorkerBase(AbstractWorker):
             if screen_type in [ScreenType.GAMEDATA, ScreenType.CONSENT, ScreenType.CLOSE]:
                 logger.warning('Error getting Gamedata or strange ggl message appears')
                 self._loginerrorcounter += 1
-                if self._loginerrorcounter < 3:
+                if self._loginerrorcounter < 2:
                     self._restart_pogo(True)
             elif screen_type == ScreenType.DISABLED:
                 # Screendetection is disabled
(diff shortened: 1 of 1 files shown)


mapadroid/worker/WorkerBase.py
```diff
@@ -637,13 +637,14 @@ def _ensure_pogo_topmost(self):
                 self._reboot()
                 break
 
-            if self._loginerrorcounter == 2:
+            if self._loginerrorcounter > 1:
```
domizei385: Doesnt this contradict line 612. Reboot after 2+ failures, but restart until 2 failures?!

crhbetz: I don't see it. First two failures restart pogo, for the third+ we go forward the extended routine.
Could you explain further what you mean?

Grennith:
Good catch, thanks!

mapadroid/worker/WorkerBase.py
Grennith: I agree with @crhbetz.
The alternative would be to not reset the errorcount and then use modulo 2 rather than checking the direct value for equality

domizei385: @crhbetz What I was getting at: When *_loginerrorcount* reaches 2, the code in lines 612-613 triggers a pogo restart. This is immediately followed by a reboot in the block starting on line 640.
This was already the case before your PR, but IMHO this is still an unnecessary step.

crhbetz: Got it now, I guess you're right and I changed one more line accordingly :-)

PR #779, opened 2020-03-21
add 0.169.3 64bit addresses + rem. older addresses

nepixl:
* removed addresses older than 0.169.0 (forced version)
- will add 32bit if available


krzee:
used this and added 32bit. thank you

PR #797, opened 2020-04-03
Updated possible dittos list

cyphercris:
Update possible dittos list

### DIFF
--- a/mapadroid/db/DbPogoProtoSubmit.py
+++ b/mapadroid/db/DbPogoProtoSubmit.py
@@ -134,7 +134,7 @@ class DbPogoProtoSubmit:
             # initialize to not run into nullpointer
 
         # ditto detector
-        if pokemon_data.get("id") in (13, 46, 48, 163, 165, 167, 187, 223, 273, 293, 300, 316, 322, 399) and \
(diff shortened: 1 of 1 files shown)


JabLuszko:
Those are event pokemon with different behavior than "normal" Dittos - we do not know if this is temporary or not (especially that Voltorb should go shiny in May, so deleted from that list). I don't think anyone is having any problems in finding Dittos now with increased spawns and million of those. I would wait for confirmation that those still can be Dittos after the event ends.

cyphercris:
So can be delete after event :dagger:

ccev:
They seem to still be be Ditto disguises.

     SELECT * FROM pokemon WHERE pokemon_id = 590 AND weather_boosted_condition > 0 AND (individual_attack < 4 OR individual_defense < 4 OR individual_stamina < 4 OR cp_multiplier < .3) ORDER BY disappear_time desc limit 10;

This query to detect possible Foongus Dittos definitely returns recent results for me. The most recent one from 5 hours ago.

Additionally, Venonat has its shiny released today (1PM local), so it stops being a possible disguise and should be removed from the list.

cyphercris:
Remove Voltrob

cyphercris:
Also Skitty should be remove

JabLuszko:
273 aka Seedot should be gone too :)

PR #806, opened 2020-04-11
Adding web socket adress to stdout in dummy_debug.py

loewe:
So it is easier to debug devices.


Grennith:
Didn't know a lot of people were using it :) Cheers

PR #812, opened 2020-04-19
fixed check_time_period

domizei385:


boby360:
I was having route/walker issues with a lot of devices after midnight. This PR worked 100%.

PR #814, opened 2020-04-21
Handle KO error from RGC correctly, properly handle login failure on PTC

### LINKED ISSUE #485: Processing ScreenType.FAILURE doesn't work for PTC accounts
***I'm using 3 phones and 15 atv - 3 phones and 1 atv have PTC accounts and according to logs this issue affects only these devices. Devices with Google accounts are fine***

So it looks like Processing ScreenType.FAILURE simply doesn't work when worker is using PTC account. Device is rebooting after processing screen failure and so on so in fact it gets stuck in a loop.

I've noticed that this loop lasts until I manually press "TRY A DIFFERENT ACCOUNT" - after that MAD simply log back in PTC account. I have confirmed that solution multiple times on all 4 devices using PTC accounts.

So my request is to make MAD press "TRY A DIFFERENT ACCOUNT" when ScreenType.FAILURE appears. This solution could be available only if `logintype: ptc` but it also should work for Google accounts.

crhbetz:
1) Identify the KO error from RGC as screentype.ERROR instead of getting wrongly caught as screentype.CLOSE in ``` elif "com.nianticlabs.pokemongo" not in topmost_app:```

2) fix #485 - on PTC the text below the green button on the failure screen needs to be clicked to be able to log in again. Independently found this issue and implemented the same fix @cecpk proposed in #485 .

I tested these for a couple of days, but cleaned them up today for the PR, so one or two tests or a couple more days in my setup would be good to 100% confirm them working.

### DIFF
--- a/mapadroid/ocr/screenPath.py
+++ b/mapadroid/ocr/screenPath.py
@@ -248,8 +248,7 @@ class WordToScreenMatching(object):
         elif screentype == ScreenType.PTC:
             return self.__handle_ptc_login()
         elif screentype == ScreenType.FAILURE:
-            self.__handle_returning_player_or_wrong_credentials()
-            screentype = ScreenType.ERROR
+            screentype = self.__handle_failure_screen(diff, global_dict)
         elif screentype == ScreenType.RETRY:
             self.__handle_retry_screen(diff, global_dict)
         elif screentype == ScreenType.WRONG:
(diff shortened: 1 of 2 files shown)


mapadroid/ocr/screenPath.py
```diff
@@ -403,6 +405,20 @@ def __handle_ptc_login(self) -> ScreenType:
         time.sleep(50)
         return ScreenType.PTC
 
+    def __handle_failure_screen(self, diff, global_dict) -> None:
```
Grennith, line 408: Returning None does not really match ScreenType ;)
`-> ScreenType:`

mapadroid/ocr/screenPath.py
```diff
@@ -129,7 +129,10 @@ def __evaluate_topmost_app(self, topmost_app: str) -> Tuple[ScreenType, dict, in
         returntype: ScreenType = ScreenType.UNDEFINED
         global_dict: dict = {}
         diff = 1
-        if "AccountPickerActivity" in topmost_app or 'SignInActivity' in topmost_app:
+        if "KO: Could not read topmost app. Is screen off?" in topmost_app:
+            logger.error("RGC could not read topmost app. Is screen off?")
+            return ScreenType.ERROR, global_dict, diff
```
Grennith, line 134: We could probably just check for startsWith "KO:" since any "KO" really is an issue. Then again, that should probably be located in the Communicator itself :thinking:

mapadroid/ocr/screenPath.py
crhbetz: Yeah, that happened during the "cleanup" :-)
Will be done in cleanup part 2.

crhbetz:
With `communicator.topmost_app()` now returning `None` when receiving a `KO:` error, the problem should be handled here:

https://github.com/Map-A-Droid/MAD/blob/595dc299fe8a26ae4401a8990d1f76a0e4dff600/mapadroid/ocr/screenPath.py#L443-L446

Better, I guess? :-)

PR #815, opened 2020-04-21
german quest task for new quest

Micha854:


sn0opy:
There's a typo in `CHALLENGE_BUDDY420_PLAY`. I assume it should be "ein Bonbon", right?

Micha854:
"ein Bonbon" or "einen Bonbon" both should be correct

sn0opy:
You're correct. What's the ingame text? We should stick to the ingame text eventhough I would never use "einen" or "der Bonbon".

Micha854:
the original text "Verdiene 1 Bonbon, indem du mit deinem Kumpel spazieren gehst"

I find it a bit long, so I cut it down to the essentials...

sn0opy:
Well then either stick to the original nor matter how long it is or use the more common used "ein Bonbon" or even "1 Bonbon" for the translation please.

Micha854:
the previous commit:
"CHALLENGE_TRICKSTER2020_FEED_BERRY": "Gib deinem Kumpel einen Snack"

have an original quest task "Gib deinem Pokemon einen Snack"

then just close the PR and I'll use the change for myself 👍

sn0opy:
I Have neither seen nor reviewed the other commit nor am I aware of the actual German texts in the game, but if we add them they should be the correct ones or at least don't cause questions because of a rarely used form eventhough it is accepted by the Duden... I wouldn't accept any commit with "kucken" either...

PR #821, opened 2020-04-22
Add framework for a utilities page

wraythex:
I've added a basic framework for a utilities page to do some maintenance on the db.  I'd like to see us add some functionality like delete quests from today that were scanned before time X, or delete pokestops that haven't been scanned in the last X days, etc.

I've also added a function that can be used for the delete quests page, but I haven't tied it in yet.  I'm new to using flask so I wanted to ask if my approach seemed good before trying anything more complex.

### DIFF
--- a/mapadroid/db/DbWrapper.py
+++ b/mapadroid/db/DbWrapper.py
@@ -1258,7 +1258,7 @@ def delete_quests_before_time(self, before_timestamp=None, from_fence=None, dele
             "WHERE GUID in ({})".format(str(','.join(quest_ids)))
         )
         self.execute(query, commit=True)
-        logger.debug("dbWrapper::delete_quests_before_time - deleted {} quest(s)".format(len(quest_ids))
+        logger.info("dbWrapper::delete_quests_before_time - deleted {} quest(s)".format(len(quest_ids)))
 
     # return number of quests matching criteria
     return len(quest_ids)
--- a/mapadroid/db/DbWrapper.py
+++ b/mapadroid/db/DbWrapper.py
@@ -344,6 +344,39 @@ class DbWrapper:
 
         return questinfo
 
+    def delete_quests_before_time(self, before_timestamp=None, from_fence=None, delete_quests=False):
+        """
+        Delete all quests before the given timestamp (UTC) from the given fenced in area.
+        Return number of matching (delete_quests=False) or deleted (delete_quests=True) quests.
+         - before_timestamp is a UTC timestamp: 1587414020 (or similar), None means delete all quests from today
+         - from_fence is a string in format: "lat1 lon1, lat2 lon2, lat3 lon3, lat1 lon1", None means delete from all areas
+           note: no comma between lat lon, and first point equals last point
+        """
+
+        quest_info = self.quests_from_db(fence=from_fence)
+
+        quest_ids = []
+        for q in quest_info:
+            if before_timestamp != None:
+                if int(quest_info[q]['quest_timestamp']) < int(before_timestamp):
+                    quest_ids.append( '"' + quest_info[q]['pokestop_id'] + '"')
+            else:
+                quest_ids.append( '"' + quest_info[q]['pokestop_id'] + '"')
+
+        logger.info("dbWrapper::delete_quests_before_time - found {} quest(s)".format(len(quest_ids)))
+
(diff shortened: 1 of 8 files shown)


wraythex:
I've added some extra functionality and believe this PR is ready for someone else to test.
Some of the things I've added are:
A new "Utilities" option under the "System" menu
A placeholder page for stops/gyms/pokemon for future utilities
A simple but functioning "Quests" maintenance page, with it you can:
 - Choose a "pokestops" area ("All" is also an option)
 - Choose an action for quests in that area (count or delete, count is default)
 - Choose a "Before Time", to only affect quests before that time (default is 8am, since that is a common reset time)
 - Choose a "Time Zone" (it should auto detect your local timezone, but you can change it if you want)
 - See the results of your requested area/action/time

muckelba:
Nice PR! Not sure how this works on a technical level, but can you add an API function as well? So we don't have to "emulate" MADmin webrequests for automating that stuff?

wraythex:
I've never seen an official API list of functions or how to use them.  But, an API way to do this would be to request the following url:
<yourserverurl>/utilities/util_q
You can use 3 arguments with it:
fence
action
beforetime

- fence needs to be the name of one of your geofences (the names inside the square brackets [area1]) or the word all which will affect all your geofences.  An example looks like fence=all or fence=area1 (defaults to all when not specified)

- action needs to be either count or delete (defaults to count when not specified)

- beforetime (bt for short) needs to be a UTC timestamp from today in your timezone.  If bt is a timestamp from before today, then no quests will be affected.  If bt is a timestamp from some day in the future, then all current quests will be affected.  Since a common reset time is 8am local, you'll need to calculate the equivalent UTC timestamp to only affect quests from before that time.  If bt is not specified, then all quests before the current time will be affected.

This API function requires authorization to use.  I'm not quite sure what ""emulate" MADmin webrequests" means, but what I'm specifying here is similar to what I've seen others suggest for other API tasks.

Some example urls look like:
<yourserverurl>/utilities/util_q
(this will return a "count" of quests from "all" areas for "all" times today)

<yourserverurl>/utilities/util_q?fence=area1
(this will return a "count" of quests from "area1" for "all" times today)

<yourserverurl>/utilities/util_q?fence=area1&action=delete
(this will delete all quests from "area1" for "all" times today)

<yourserverurl>/utilities/util_q?fence=area1&action=delete&beforetime=1587414020
(this will delete all quests from "area1" from before UTC 1587414020, if that was during your current day.  since now it will be from before your current day, it won't affect any quests)

muckelba:
> I've never seen an official API list of functions or how to use them. 

https://mad-docs.readthedocs.io/en/latest/api/

> This API function requires authorization to use. I'm not quite sure what "emulate" MADmin webrequests" means

just firing up `curl` commands instead of clicking buttons in MADmin
I mean that's fine and all, but using a "real" API has few advantages like input validation and usable returning data.

wraythex:
That page starts to explain the API, but does not go into full details.  If the mad developers would like to move the util_q function from /utilities to /api, that can be done and it would still work the same.  If the mad developers are fine with the function where it is, then my description above can stand as the API description for the function.

muckelba:
cc @Expl0dingBanana . He built the API :) As i said, the current implementation is fine, but it's kinda inconsistent when almost everything else is usable with the API

wraythex:
Ah, I see how clicking through gives more info for the api.  However, like my first post said, I'm still new to flask and not very familiar with where everything in mad is located or should be placed.  I created this 10 days ago and haven't had any feedback on this PR.  I know people are busy and have other priorities, so I was just waiting to hear back from users for functionality or developers for code review or comments.

Expl0dingBanana:
I agree with @muckelba we should have the functionality in the API.  After reviewing the calls it doesnt look too hard to migrate them over to the API.  Take a look at how APKs are being handled (as they have a relatively simple implementation)

https://github.com/Map-A-Droid/MAD/tree/master/mapadroid/madmin/api/apks

wraythex:
I've tried to look over the api code, but I'm not familiar with it enough to port this code over.  I can close this PR if any devs want it closed.  Or maybe we can have it stand as a request for a utilities page that can do quest/stop/gym/route maintenance?  Devs, let me know what you'd like me to do with this PR.

wraythex:
I'll try to revisit this in the future.  Closing for now.

PR #825, opened 2020-04-23
Improve distance to check against

crhbetz:
Make sure to have proper distances for the recently introduced distance checks.
Could fix issues with quest / level / iv_mitm modes.

### DIFF
--- a/mapadroid/worker/MITMBase.py
+++ b/mapadroid/worker/MITMBase.py
@@ -58,6 +58,7 @@ class MITMBase(WorkerBase):
         max_radius = self._mapping_manager.routemanager_get_max_radius(self._routemanager_name)
         if not max_radius:
             return True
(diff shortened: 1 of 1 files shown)


sn0opy:
Without looking super deep into the code:

shouldn't we change the check in some way to ignore actual GMO data but keep everything else? An encountered mon is still an encountered mon. There's no way the worker hasn't been at the place where the mon is to encounter it.

Grennith:
> Without looking super deep into the code:
> 
> shouldn't we change the check in some way to ignore actual GMO data but keep everything else? An encountered mon is still an encountered mon. There's no way the worker hasn't been at the place where the mon is to encounter it.

The check only every runs on GMOs. Since scanning is based on a route, reaching our destinations is of importance.

PR #829, opened 2020-04-25
Update the trs_status schema for older installs

Expl0dingBanana:
The column lastProtoDateTime had 'on update current_timestamp()'.  This would cause the table to be updated for the device even when no data was sent.  This field is set when data is correctly recieved from the worker so no additional changes are required


Expl0dingBanana:
Looks like when I rebased this I lost the actual commit.  Closing this and will open a new PR when fixed

PR #833, opened 2020-04-28
WIP: accept and parse raw protos

sn0opy:
Work in progress. Do not use. Do not merge.

This change makes MAD and PD a bit more future-proof. Right now, MAD only works with the parsed JSON data sent from PD. These protos needs to be updated by the core team with every major change to the protos within the game which luckily are quite rare.

PD is already able to send raw protos, but MAD simply ignores it. This is going to change with this PR. Every message has a `raw` boolean field. If true, the protos will be parsed on the server's end instead of the device.

Simply implementing this doesn't work. The Python protos we use have a different schema compared to the ones we use in PD's serialized output. This means we need to map them to an equal schema first by either implementing a completely new schema and use it everywhere or map parsed protos to the PD schema we have right now.

I was looking into SQLAlchemy to map both to an equal new schema + benefit from the clean ORM stuff which should also clean up some dirty SQL  we have.

Todos:
- [x] add protos
- [x] use protos
- [x] create new models
- [ ] map serialized and parsed data to new models
- [ ] implement SQLAlchemy for game data first 

Pros:
- Parsing data on the device takes a little bit of CPU and RAM. Since we cannot upgrade devices moving the parsing part to a server makes sense
- There are rumors, that people scan with iOS devices which send data in raw. MAD now supports raw data. Just do the math.
- Updating protos on the server's end is quicker than waiting for an app update and rolling out said update
- a PD update is not required for this change

Cons: 
- this PR changes/adds 1300 files. Chrome doesn't really like that when viewing the commit on Github
- make sure you can upgrade your servers. RaspberryPI scanning might come to an end with this


krzee:
i love this idea

Expl0dingBanana:
Discussed with @sn0opy and we will close this PR for now but leave the branch open for development.

PR #846, opened 2020-05-10
Separate quest/stop stats

cecpk:
Separate Quest / Stop in own statistic page 
image

**New feature**
quest process status
image


moltova:
works fine for me on desktop & mobil, thx

PR #862, opened 2020-05-18
Fix SQL statement to use index instead of table scan

DrMurx:
Doing any kind of calculation on an indexed column prevents MySQL
from using the index for queries.

So instead of

```
WHERE function(indexedcolumn) = expression
```

doing

```
WHERE indexedcolumn = inversefunction(expression)
```

can improve query performance in order of magnitudes.


DrMurx:
Effect on the CPU usage of MariaDB (light blue line) going from 300/400% down to less than 100%.

image

PR #868, opened 2020-05-22
MAD APKs storage interface

Expl0dingBanana:
Allows MAD to save files to the filesystem or database depending on the option selected.  This will default to filesystem and can be changed with the flag -sdb.


mapadroid/utils/walkerArgs.py
```diff
@@ -39,6 +39,7 @@ def parseArgs():
                         default=os.getenv('MAD_CONFIG', os.path.join(mapadroid.MAD_ROOT,
                                                                      'configs/mappings.json')),
                         help='Set mappings file')
+    parser.add_argument('-sdb', action='store_true', default=False, help='Save MAD APKs to database')
```
fosJoddie, line 42: I got a couple of comments about this setting
- SDB - what is it short for?
- missing a long name

I propose to change the name to `asi`  short for `apk-storage-interface` - choice of two: `fs` and `db` with a default of `fs`

Expl0dingBanana: Code has been updated to reflect asi / apk_storage_interface

Expl0dingBanana:
RunTimeWarning should no longer appear as that does not affect the download.

Issue #664, opened 2020-01-07
Prevent devices from rebooting


kamieniarz:
So basically when reboot device is false, devices can still reboot (eg when they can't inject in time). It would be nice if there would be an option to completly prevent mad from rebooting devices.

LucianoNeo:
Still waiting for the fix...

Expl0dingBanana:
Please test PR 871

Expl0dingBanana:
@kamieniarz @LucianoNeo can either of you validate PR 871?

kamieniarz:
I reported this issue 4,5 months ago so unfortunately no, I already gaved up on some settings

LucianoNeo:
Same here, just discarded the phones that can't reboot...

crhbetz:
I can confirm #871 working. Had multiple devices loop up to 100+ counts of  `not injected yet` with `reboot` set to `False` and `injection_thresh_reboot` at default/empty:

`[           mini01] [           MITMBase:261 ] [    INFO] PogoDroid on worker mini01 didn't connect yet. Probably not injected? (Count: 103/Unlimited)`

PR #871, opened 2020-05-24
Device will only reboot during injection if reboot == true

Expl0dingBanana:
Fix for issue #664


Expl0dingBanana:
Closing this is the feature is no longer requested.

@crhbetz thanks for providing the feedback

PR #883, opened 2020-05-29
Add Link to Item IDs

spammer23:
Maybe it would be nicer to have an real clickable Link there..


spammer23:
Just found the dropdown..
Not sure if needed. Can be closed

cecpk:
we dont need that :D

PR #887, opened 2020-06-03
Add OR-Tools routecalc option

wraythex:
This PR creates a new file which will use the OR-Tools for routecalc's if it is available.  If not, it will return the original requested (quick/optimized) MAD routecalc.  I believe this will be a big benefit to MAD as the OR-Tools are very fast at calculating routes, and the length of the path found (for this test data) was always shorter than what MAD found.

Here are some runtime and route length comparisons I've run (all on a single core of my home computer) for various number of coordinates:
coords  mad   vs ortools :   mad_len  vs   or_len   : or_len/mad_len
   50    1.5s vs   0.04s : 0.50242212 vs 0.50087158 : 0.996913
  100   14.4s vs   0.22s : 0.54041637 vs 0.53329480 : 0.986822
  150   26.4s vs   0.46s : 0.57627646 vs 0.57385556 : 0.995799
  200   54.8s vs   0.96s : 0.69595905 vs 0.68024545 : 0.977421
  250  148.5s vs   2.12s : 0.78786951 vs 0.77921291 : 0.989012
  300  229.4s vs   3.08s : 0.90786359 vs 0.87315918 : 0.961773
  350  397.8s vs   2.89s : 1.07814378 vs 1.06397047 : 0.986853
  400  674.1s vs   4.29s : 1.28250018 vs 1.23183531 : 0.960495
  450  554.9s vs   6.16s : 1.51514236 vs 1.44230409 : 0.951926
  500  812.1s vs  10.44s : 1.81889609 vs 1.72162878 : 0.946523
  930 6416.1s vs  33.36s : 4.31855154 vs 4.00745554 : 0.927962

I'll be testing this PR on my installation and will report back when I believe it is stable.

### DIFF
--- a/mapadroid/data_manager/modules/routecalc.py
+++ b/mapadroid/data_manager/modules/routecalc.py
@@ -159,11 +159,8 @@ class RouteCalc(Resource):
             logger.info("Calculating a short route through all those coords. Might take a while")
             from timeit import default_timer as timer
             start = timer()
-            if algorithm == 'quick':
-                from mapadroid.route.routecalc.calculate_route_quick import route_calc_impl
-            else:
-                from mapadroid.route.routecalc.calculate_route_optimized import route_calc_impl
-            sol_best = route_calc_impl(lessCoordinates, route_name, num_processes)
+            from mapadroid.route.routecalc.calculate_route_all import route_calc_all
+            sol_best = route_calc_all(lessCoordinates, route_name, num_processes, algorithm)
             end = timer()
             logger.info("Calculated route in {} minutes", str((end - start) / 60))
             for i in range(len(sol_best)):
--- a/mapadroid/route/routecalc/calculate_route_all.py
+++ b/mapadroid/route/routecalc/calculate_route_all.py
@@ -1,0 +1,118 @@
+import math
+import secrets
+
+from mapadroid.utils.logging import logger
+from .util import *
+
+try:
+    from ortools.constraint_solver import routing_enums_pb2
+    from ortools.constraint_solver import pywrapcp
+except:
+    pass
+
+
+def create_data_model(lessCoordinates):
+    """Stores the data for the problem."""
+
+    data = {}
+
+    # ortools requires x,y data to be integers
+    # we will scale lat,lng to large numbers so that rounding won't adversely affect the path calculation
+    data['locations'] = []
+    for e in lessCoordinates:
+        data['locations'].append( ( int(float(e[0])*1e9), int(float(e[1])*1e9) ) )
+
+    data['num_vehicles'] = 1 # calculate as if only one walker on route
+    data['depot'] = 0 # route will start at the first lat,lng
+    return data
+
+
+def compute_euclidean_distance_matrix(locations):
+    """Creates callback to return distance between points."""
+    distances = {}
+    for from_counter, from_node in enumerate(locations):
+        distances[from_counter] = {}
+        for to_counter, to_node in enumerate(locations):
+            if from_counter == to_counter:
+                distances[from_counter][to_counter] = 0
+            else:
+                # Euclidean distance
+                distances[from_counter][to_counter] = (int(
+                    math.hypot((from_node[0] - to_node[0]),
+                               (from_node[1] - to_node[1]))))
+    return distances
+
+
+def format_solution(manager, routing, solution):
+    """Format the solution for MAD."""
+    route_through_nodes = []
+    index = routing.Start(0)
+    while not routing.IsEnd(index):
+        route_through_nodes.append(manager.IndexToNode(index))
+        index = solution.Value(routing.NextVar(index))
+    return route_through_nodes
+
+
+def route_calc_ortools(lessCoordinates, route_name):
+
+    data = create_data_model(lessCoordinates)
+
+    # Create the routing index manager.
+    manager = pywrapcp.RoutingIndexManager(len(data['locations']),
+                                           data['num_vehicles'], data['depot'])
+
+    # Create Routing Model.
+    routing = pywrapcp.RoutingModel(manager)
+
+    distance_matrix = compute_euclidean_distance_matrix(data['locations'])
+
+    def distance_callback(from_index, to_index):
+        """Returns the distance between the two nodes."""
+        # Convert from routing variable Index to distance matrix NodeIndex.
+        from_node = manager.IndexToNode(from_index)
+        to_node = manager.IndexToNode(to_index)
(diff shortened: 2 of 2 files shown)


sn0opy:
How do you read the "length"? Is it the total length in meters? 

Do you happen to have some comparison screenshots of these routes? 

Have you compared both route algorithms with yours as well?

wraythex:
I calculated the length just as the Euclidean distance between each point in the path, not considering geography at all.  This wouldn't be good to compare distances between sets of data at different North/South latitudes, but for one data set in one location, it should work well.

Here is the spread of distances for the 200 coordinates case:

Distance Bins | Mad | Ortools
0 | 0 | 0
0.001 | 124 | 128
0.002 | 34 | 30
0.003 | 11 | 12
0.004 | 5 | 3
0.005 | 4 | 5
0.006 | 4 | 3
0.007 | 4 | 5
0.008 | 1 | 2
0.009 | 0 | 1
0.01 | 1 | 0
0.011 | 2 | 2
0.012 | 1 | 0
0.013 | 1 | 0
0.015 | 0 | 1
0.019 | 2 | 3
0.026 | 1 | 1
0.027 | 1 | 0
0.034 | 1 | 1
0.114 | 1 | 1
0.164 | 1 | 1

Where each distance bin (d) represents a range of distances between [d-0.001, d], and the other two columns report how many times Optimized Mad or OR-Tools had a route length in that bin.  I removed rows where both tools had 0 entries.  If you have another way you'd like to see the data, please let me know.  If you'd like to see screenshots of routes, I'll have to calculate it with different data since this original analysis was done with local data.

I have not compared the Mad Quick Calc vs the OR-Tools.  Would you like to see that as well?  I thought that the Mad Optimized Calc would give the best distance comparison to the OR-Tools, but I could understand if you'd like to see how the Quick compares to the OR-Tools as well.  Let me know.

sn0opy:
> I have not compared the Mad Quick Calc vs the OR-Tools. Would you like to see that as well? I thought that the Mad Optimized Calc would give the best distance comparison to the OR-Tools, but I could understand if you'd like to see how the Quick compares to the OR-Tools as well. Let me know.

Well, a visual and time based comparison would be interesting between both of them, since OR-tools  adds a) new dependencies and b) only seems to work on 64 bit. We don't have any statistics for the MAD server architecture people use, but yeah. who knows.

All in all it looks pretty nice and I really like the idea of having just falling back from one method to another.

wraythex:
I pulled 300 "random" coordinates from Paris and ran the calculations on those.  Here are the results:

  | Length | Time
-- | -- | --
In-Order | 5.1787133 | 0.000s
Quick | 0.3206942 | 0.341s
Optimized | 0.3100333 | 271.163s
Ortools | 0.2856116 | 2.495s
Or/Quick | 0.890604 |  
Or/Optimized | 0.921228 |  

Where In-Order is just the list of coordinates returned from the database.  I'm including images for each of the 4 routes listed above.  In-Order zips left and right, from bottom to top across the coords.  Quick has several criss-cross routes.  Optimized had only one criss-cross route.  Ortools had no criss-cross routes.

Here is the distribution of Euclidean lengths between points with a bin size of 0.0002:

Bins | Quick | Optimized | Ortools
-- | -- | -- | --
0 | 0 | 0 | 0
0.0002 | 1 | 1 | 1
0.0004 | 25 | 26 | 26
0.0006 | 49 | 48 | 51
0.0008 | 68 | 60 | 69
0.0010 | 50 | 46 | 59
0.0012 | 30 | 29 | 21
0.0014 | 18 | 30 | 24
0.0016 | 15 | 18 | 16
0.0018 | 10 | 10 | 9
0.0020 | 5 | 10 | 7
0.0022 | 4 | 5 | 7
0.0024 | 5 | 3 | 3
0.0026 | 5 | 1 | 0
0.0028 | 1 | 3 | 1
0.0030 | 3 | 2 | 2
0.0032 | 1 | 1 | 0
0.0034 | 2 | 0 | 0
0.0036 | 1 | 2 | 1
0.0038 | 1 | 0 | 0
0.0042 | 2 | 1 | 0
0.0046 | 0 | 0 | 1
0.0048 | 0 | 2 | 0
0.0054 | 0 | 1 | 0
0.0070 | 0 | 0 | 1
0.0074 | 1 | 0 | 0
0.0076 | 2 | 0 | 0

I still haven't had a chance to test the PR out yet.  I'll let you know when I have.

route0_in_order
route1_quick
route2_optimized
route3_ortools

sn0opy:
All right, all in all things are looking really good. I will ask around if people experience any bigger issues with the quick method we have right now. 

I think we could actually think about removing the ability to choose a calculation method at all. While doing so, we replace the existing optimized mode with your ortools implementation and fall back to the existing quick mode if requirements aren't met.

sn0opy:
Been getting some feedback from others. I think we do exactly that. We get rid of our existing optimized route calculation.

@wraythex Are you interested in working on that one as well? I might take over otherwise.

1. remove ability on frontend or anywhere else to choose the calculation method
2. basically use your implemented checks but remove the fallback to the optimized mode

wraythex:
I've tested this PR on my server and it worked.

I first made sure route recalc worked with master.  It did.
(optimized algo on a 94 length quest route, single geofence with two [named] areas, ~30 seconds)
I then switched to pr_887 (without OR-Tools installed).
A route recalc worked (it used the optimized algo, ~30 seconds).
I then stopped mad, installed OR-Tools with pip in my venv, and started mad.
This route recalc used ortools and took about 0.5 seconds.

I think I'd like to leave the removal of the optimized algorithm up to you.  I see that it is referenced in the db, and in a few places in the code.  I'd hate to miss anything.

I think we could make some improvements to the existing quick calc code, perhaps by trying to remove, or reduce, the number of criss-crosses in a path.  But I'm not in a position to work on that right now.  I may be able to look at that more in the future and open a new PR then.

Is there anything else I can do for this PR?

sn0opy:
Nah, it's fine. I'll merge this into a feature branch and rework the route calculation stuff to basically switch between yours and the existing quick mode. There's not really a reason to look into the way the quick mode works right now. People should use the ortools mode to get a proper optimized route. Will look into it on the weekend.

ChaosEmpire:
maybe the routefree mode could use or-tools too

sn0opy:
> maybe the routefree mode could use or-tools too

the idea of the routefree mode is not to generate a fixed route but find the closest stop to the current position.

ChaosEmpire:
was just an idea, to speed also this process up, its not that long, but happens several times..

"Kleinvieh macht auch Mist"

PR #908, opened 2020-06-16
fix a mix-up of ScreenTypes FAILURE and RETRY

crhbetz:
See title. Also:

+ remove the handling procedure that was wrongly implemented because
of this mix-up.
+ describe the ScreenTypes more clearly.

This is ScreenType.FAILURE - one green button saying OK. Obviously we want to click the button here:
grafik

This is ScreenType.RETRY - green button on top saying RETRY, clickable "non-button text" below saying "TRY A DIFFERENT ACCOUNT". We want to click the "non-button text":
grafik

Edit: Due to the nature of the issue - dealing with an error condition that has to pop up first - this is untested beyond making sure that MAD still starts.


acocalypso:
thanks, Ill test it out and report back.

acocalypso:
after testing it over night, I had my first questscan which completed successfully, almost all devices were up and running.

PR #911, opened 2020-06-16
Add origin/device to log line

JabLuszko:
Not doing anything more, waiting for @Expl0dingBanana rework part 2 :-)

### DIFF
--- a/mapadroid/route/RouteManagerBase.py
+++ b/mapadroid/route/RouteManagerBase.py
@@ -633,7 +633,7 @@ class RouteManagerBase(ABC):
                     return self.get_next_location(origin)
                 self._last_round_prio[origin] = True
                 self._positiontyp[origin] = 1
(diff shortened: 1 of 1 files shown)


Expl0dingBanana:
Can you try the branch origin_logger_rework?

JabLuszko:
There is routename (aka Area) not devicename.
I would prefer device name for "grep <devicename>" to see full flow, but I kinda get the routename there too.
I guess we need both there then!

Expl0dingBanana:
This has been implemented in d7ead1f by using the routemanager logger.  This logger will use the origin as the identifier and add the prefix '{RouteManager Name}:' to any message.  Below is an example log line.
`[06-24 11:37:12.85] [         devbox01] [I] [Raids]WhiteHouse-DC-USA: Moving on with location 38.88963758338736, -77.02668313499373 [1 coords left (Workerpool)]`

PR #913, opened 2020-06-18
Device will only reboot during injection if reboot = true

Expl0dingBanana:
Continuation of #871


LucianoNeo:
Working fine for 2 days without reboot!

PR #920, opened 2020-06-24
Remove weedle from ditto

clburlison:


cecpk:
rly?

krzee:
> 
> 
> rly?

nothing which has a shiny can be ditto, we just had weedly community day

PR #927, opened 2020-07-10
Madmin logcat download

Grennith:
RGC can now send back a logcat compressed in a ZIP file.
In order to get the logcat, I added a button to the devicecontrol template.
If pressed, MAD sends the command to RGC, receives the ZIP file, stores it in the temp folder and directly sends it off to the user. The user just has a brief delay with a download of logcat_{origin}.zip


mapadroid/madmin/functions.py
```diff
@@ -113,6 +113,11 @@ def generate_device_screenshot_path(phone_name: str, device_mappings: dict, args
     return os.path.join(args.temp_path, screenshot_filename)
 
 
+def generate_device_logcat_zip_path(origin: str, args: dict):
+    screenshot_filename = "logcat_{}.zip".format(origin)
```
na-ji, line 117: it's not a screenshot

Grennith: good old copypasta :)

PR #933, opened 2020-07-19
Fix statistics GROUP BY standard violations

Artanicus:
Now compatible with ONLY_FULL_GROUP_BY which has been the default and
heavily recommended since MySQL 5.7. Fixed in a way that is also compatible with MariaDB.

+ some capitalization fixes for queries just so I could inspect them easier.

### DIFF
--- a/mapadroid/db/DbSanityCheck.py
+++ b/mapadroid/db/DbSanityCheck.py
@@ -8,7 +8,7 @@ logger = get_logger(LoggerEnums.database)
 
 
 class DbSanityCheck:
-    blacklisted_modes = "NO_ZERO_DATE NO_ZERO_IN_DATE ONLY_FULL_GROUP_BY"
+    blacklisted_modes = "NO_ZERO_DATE NO_ZERO_IN_DATE"
 
     def __init__(self, db_exec: PooledQueryExecutor):
         self._db_exec: PooledQueryExecutor = db_exec
--- a/mapadroid/db/DbStatsReader.py
+++ b/mapadroid/db/DbStatsReader.py
@@ -16,14 +16,14 @@ class DbStatsReader:
     def get_shiny_stats(self):
         logger.debug3('Fetching shiny pokemon stats from db')
         query = (
-            "SELECT (select count(DISTINCT pokemon.encounter_id) from pokemon inner join trs_stats_detect_mon_raw on "
-            "trs_stats_detect_mon_raw.encounter_id=pokemon.encounter_id where pokemon.pokemon_id=a.pokemon_id and "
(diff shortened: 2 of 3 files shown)


Expl0dingBanana:
@DrMurx do you have any issues with requiring 'ONLY_FULL_GROUP_BY'?

@Artanicus can you fix the merge conflict?

Artanicus:
Conflict fixed.

To clarify, these changes do not mean ONLY_FULL_GROUP_BY must be enabled, rather that the queries work even if it is enabled.

DrMurx:
> @DrMurx do you have any issues with requiring 'ONLY_FULL_GROUP_BY'?

I'm not really having a stake at that code (just refactored it out of the humongous `DbWrapper`). But I'm absolutely for standard compliance, so doing aggregate queries properly is 👍👍👍.

However, removing `ONLY_FULL_GROUP_BY` from the SQL Mode blocker means that for development and testing, this mode **has** to be enabled to spot flawed queries in the future before they're merged into production code.

Artanicus:
@DrMurx Good point. It could be added as a recommendation for developers but not everyone will enable it since if they run any non-compliant software off the same DB server you can't really. In some ways it's up to code review to catch faulty GROUP BY statements but it's not easy for us dinosaurs used to the undefined behavior of years past. Ultimately it might fall on users to catch when it breaks since many new users will have it enabled.

TL;DR; multi-pronged approach is probably the best:
- Amend dev guide to recommend enabling ONLY_FULL_GROUP_BY to catch undefined behavior
- Write tests for the CI to catch it. (Are there even db tests now? Should probably look into it.)
- Try to catch them in code review where the tests can't.
- New users will surely report broken statistics eventually if all else fails.

Expl0dingBanana:
@Artanicus 
Are you currently running this on your system without issues?

Artanicus:
I'm running this without issues. The only places where GROUP BY is used is in the statistics and once in the DBWrapper. The wrapper use-case is dead simple and was already compliant, I just did cosmetic fixes there. For statistics everything works fine that I am actively using but will send out a Mon scanner now to see if those stats also work.

Artanicus:
Can confirm, all statistics views seem to be working normally. (100% IV and shiny views I can't confirm since just enabled Mon scanning.)

JabLuszko:
Does your "per device" stats: MADmin -> Statistics -> Global stats -> (click device name) works?
It's broken here due to `MIN()` in group by and complains from MySQL: ERROR 1111 (HY000): Invalid use of group function

https://github.com/Map-A-Droid/MAD/commit/42e07f2c5246085035ed7203b3fe4d3f7884804e#diff-157e56d1cbb5da5e4ac3b9ef2f372a04R82

PR #938, opened 2020-07-29
Wizard re-cacher

Expl0dingBanana:
Allow the wizard to re-create the cache on-demand by calling an endpoint.  This is useful for multi-instance systems so the download only needs to go through on a single instance


crhbetz:
Test looked good on my setup. I'll try it in production when we get the 183.0 addresses.

PR #941, opened 2020-07-30
rearrange spawn details from spawn statistics

JabLuszko:
Rename columns: Last scanned to Last despawn time confirmation; add new "Last mon scanned"
Truncate GPS coords to 8 chars (you can still search by full coords, this is only display thing) - I'd rather see last mon seen time rather than full 18 chars long GPS coords
Drop that strange Date() thing - I don't think we really need moment.js here for spawn times overview, this should be enough

Drop that stateSaveCallback/stateLoadCallback that @cecpk loves everywhere when we don't need it at all 
```
            "stateSave": true,
            "stateDuration": 0
```
take cares of everything.


sn0opy:
Needs a rebase. Would you like to have a look on that?

sn0opy:
@JabLuszko ☝️

PR #943, opened 2020-08-03
Fix not being able to save event after flake8

JabLuszko:
Typo.

Also no complains Travis or I kill you, thanks :D


Expl0dingBanana:
Conflict has been resolved in #944 so merging is fine

Issue #907, opened 2020-06-16
Apk Wizard cannot download Pogo apk on Raspberry Pi


Tabbomat:
Currently, the wizard cannot download the Pogo apk via  matlink-gpapi, it fails with the following error:

```
[06-16 12:10:32.56] [           system] [             wizard:260 ] [I] Searching for a new version of PoGo [arm64_v8a]
[06-16 12:10:32.96] [           system] [    gplay_connector:209 ] [I] Using cached token.
[06-16 12:10:38.88] [           system] [             wizard:270 ] [I] Newer version found on the Play Store: 0.177.1
[06-16 12:10:38.96] [           system] [    gplay_connector:209 ] [I] Using cached token.
[06-16 12:10:45.38] [           system] [    gplay_connector:62  ] [I] Starting download for com.nianticlabs.pokemongo
[06-16 12:10:46.48] [           system] [    gplay_connector:69  ] [E] Error while downloading com.nianticlabs.pokemongo : Unexpected end-group tag.
[06-16 12:10:46.49] [           system] [             wizard:145 ] [I] Issue downloading apk
[06-16 12:10:46.50] [           system] [             wizard:148 ] [W] Unable to successfully download the APK
[06-16 12:10:46.50] [           system] [    gplay_connector:209 ] [I] Using cached token.
[06-16 12:10:52.71] [           system] [    gplay_connector:62  ] [I] Starting download for com.nianticlabs.pokemongo
[06-16 12:10:53.74] [           system] [    gplay_connector:69  ] [E] Error while downloading com.nianticlabs.pokemongo : Unexpected end-group tag.
[06-16 12:10:53.75] [           system] [             wizard:145 ] [I] Issue downloading apk
[06-16 12:10:53.76] [           system] [             wizard:148 ] [W] Unable to successfully download the APK
[06-16 12:10:53.76] [           system] [    gplay_connector:209 ] [I] Using cached token.
[06-16 12:11:00.63] [           system] [    gplay_connector:62  ] [I] Starting download for com.nianticlabs.pokemongo
[06-16 12:11:01.98] [           system] [    gplay_connector:69  ] [E] Error while downloading com.nianticlabs.pokemongo : Unexpected end-group tag.
[06-16 12:11:01.99] [           system] [             wizard:145 ] [I] Issue downloading apk
```

This happens for both v7 and v8 pogo apk, but not for RGC or PD.

[...]

PR #946, opened 2020-08-07
RPi wizard fix

Expl0dingBanana:
Fixes an issue that caused RPi to be unable to download from the play store.


Expl0dingBanana:
Merging based off feedback from #907

PR #947, opened 2020-08-07
Fix jobs that run from control page

Expl0dingBanana:
* Fix an issue where jobs could not be run from the control page due to a function signature error


Expl0dingBanana:
Merging based off feedback from [discord ](https://discordapp.com/channels/465247740553592832/467318193141317659/741054751789154336) on the branch.

PR #959, opened 2020-08-19
Implemented geofence editing on map

MrJul:
This pull request implements geofence editing directly on the madmin map.

I like to precisely tweak my fences, and was previously resorting to converting the fence coordinates to GeoJSON, edit that with a proper editor, convert it back and reimport... I did that three times before I grew tired of it and started implementing this directly on the map ;)

Several things have been changed in the madmin map code:

- I first tried to use the existing `leaflet-draw` library included in MAD. However, it hasn't been updated for more than two and a half years: any leaflet version over 1.2 was causing exceptions on editing. Thus, this PR uses [`leaflet-geoman`](https://github.com/geoman-io/leaflet-geoman) (formerly known as `leaflet.pm`) instead: draw has been completely removed.

- Since an area can have one included and one excluded fence, I didn't find a great way to edit both of them at once while keeping things simple. Instead, I've renamed the existing *Geofences* tab in the left sidebar to *Areas*, which aren't editable directly and added a new *Geofences* tab, with editing support. This also better matches the settings section and API.

- Adding a new geofence now also uses geoman and reuses most of the code from the editing mode.

- Popups have been added to geofences, areas and routes. The geofence popup contains an *Edit* button that enables the edit mode on this particular fence, then switches to *Save*/*Cancel* buttons. The fence's name can also be edited directly within the popup.

- Saving a geofence in the backend is done using the existing madmin API rather than a map-specific handler.

- Displayed fence settings are saved to the local storage using IDs insteaf of names. Using IDs made name changes work seamlessly.

- Since fences can now be interacted with, ordering layers correctly was important: the user needs to be able to click on a fence, but still interact with stops, spawnpoints, etc within. This was already partly done previously using leaflet's panes. Yet in combination with `preferCanvas: true` it prevented any mouse interaction on the background layers (fences and routes). While disabling `preferCanvas` to use SVG made everything work, performance was absolutely terrible with thousands of spawnpoints. Unexpectedly, getting this right is where I spent most of the time. I found a few gists and a plugin named [`leaflet-event-forwarder`](https://github.com/danwild/leaflet-event-forwarder), but they had many limitations: they weren't working correctly with more than two layers, didn't pass the cursor changes to background layers or went in the way when dragging points. In the end, I almost rewritten this mouse handling from scratch (after playing with a single pane and reordering all layers automatically – which mostly worked but had crappy perf).

- Each time I had to modify an existing JS function, I did some code cleanup: first by ensuring that indents were consistent (4 spaces were used most of the time, but some lines had 2 spaces), as well as unifying the quote and nested member usage styles. I also simplified other things (proper `this` when possible, reusable function for data fetching, `let`/`const` instead of `var`, etc.). At the end of the day, there are more modifications than I intended to. In an ideal world all those changes would be in a separate PR, but since I made them while implementing the new feature they aren't that easy to split. Sorry about that!

Other small map improvements:

- Title attributes have been added to the sidebar tabs.

- Interaction with various layers has been made more obvious, with the border changing on mouse over.

- Gym and stop images are now correctly sized in their popup.

Two fixes were made to the backend API handler:

- `Request.mimetype` is used instead of `content_type`. Otherwise map requests were failing because axios sends `application/json; charset=utf-8` (which is perfectly valid and should pass) instead of simply `application/json`.

- HTTP code 415 (Unsupported Media Type) is sent instead of 422 (Unprocessable Entity) if the content type isn't recognized.


sn0opy:
First of all, thanks for all the work you put into it!

I had a quick look into it and I encountered a few issues while one is a major issue. Since you refactored the way the data is retrieved from the backend, the map now fetches everything all the time eventhough it has not been activated. Ontop of that, it doesn't seem to take the timestamp query parameter into affect anymore. Usually that timestamp is used to only fetch elements that have changed since then. Due to this issue several other things seem to be broken like the periodical cleanup task or even displaying mons on the map.

The geofence editing feature, the change to seperating areas + geofences and the refactoring in general is awesome.

MrJul:
Indeed, static layers were always fetched even if disabled, woops, it's fixed.

Regarding the mons, it was an exception when adding the layer, also fixed! (I had all kinds of layers enabled while testing, but no active mons in practice since it was an old non-active database, so I completely missed the bug, sorry.) Cleanup works fine after this fix.

Apart from that, I've checked that the timestamp is still correctly passed to all applicable routes (gyms, spawns, quests, stops, mons, cells). No visible changes here. But I noticed that stops, spawns and cells aren't correctly filtered by timestamp server side. This problem was already present in master and is not new in this branch. The timestamp is ignored when old coordinates are specified: https://github.com/Map-A-Droid/MAD/blob/d957bb03e715a2e4f02417025195cbd401e04c04/mapadroid/db/DbWrapper.py#L879-L885 I can try to fix that later in another PR :)

I switched a live MAD instance to the geofence-editing branch for the past half hour and noticed no problems so far.

sn0opy:
Thanks for the fixes. Had a quick look and it seems all my "complains" were fixed. I assume the fact that everything was loaded everytimes was either caused by one of the other things or it was just me being stupid and overlooked something on my test setup. 

All in all it looks good to me. There's one last thing I'd like to be changed. The event forwarder lib. Personally I'd like to have all dependencies externally provided by either cdnjs or jsdelivr. Haven't looked for it yet and I'm not sure if any of the CDNs provides it.

MrJul:
I couldn't find the original one on CDN, and I edited it so heavily that it's basically a new one now :)
In theory it should work with anything but in practice I didn't test on anything other than the madmin map.
If you prefer I can include it directly `madmin.js`.

Expl0dingBanana:
Thanks for fixing a potential API bug with this PR 👍

fosJoddie:
This PR fixes an issue where you could not have "cell updates" turned on if you wanted to send a worker to a location (click would just open cell status popup)

After running this a few days, these are my only "issues":

The hover effect on routes makes it even harder to mouse over stops for the popup.  Not a issue with this PR directly, but especially the dots of pokestops should be larger, or take advantage of the leaflet plugin "almost over" that makes hitting tiny things on maps very easy

The list of geofences is listed by how they area read from DB, should be sorted by name ideally. get_geofences is just missing the ` geofences_sorted = sorted(geofences, key=lambda x: geofences[x]['name'])` pretty much (and iterating over that instead)

MrJul:
Geofences are now sorted in the sidebar. This is done client-side (server-side wasn't enough because IDs were used as object keys anyways) and is refreshed whenever a fence is added a renamed.

Regarding the hover, that's something that was already existing and could indeed be tweaked later by changing stops or using another plugin, as you said.

Multiple polygons could also be added easily later on the map since geoman supports multiple polygons as well as makings cuts (holes) in polygons. It will probably require some backend / route manager changes though.

fosJoddie:
Awesome update, thanks for the hard work 💪

PR #977, opened 2020-08-30
Prioq nodes now have a popup again. Added prioq line popup. Fixes #976

### LINKED ISSUE #976: #959 Layer conflict with priority route information.
As of the new commit involving geofence editing within madmin, it seems that there is a conflict between layers so you can no longer get a due time on the jump.

Jab mentioned in discord that they assume it is a layer conflict.

MrJul:
Fixed priority route nodes not having a popup after #959 (fixes #976).
Also added a popup on the route polyline itself, to match the normal route one.


AboveColin:
Looks likes it works, great!
image
image

PR #984, opened 2020-09-02
Optimized mon statistics page load

MrJul:
The mon statistics page took almost 8 seconds to load for me with 780k encounter rows in the db.

Here are the changes:

1. The recent 100% mons query was the main culprit. No indexes were used for this query: the one on `individual_attack` apparently wasn't sufficient. After adding on index on the 3 IV stats, the query is almost instant. I've also deleted the attack index since it's covered by the new one.

2. `pdfmake` has been updated; for some reason the older version was slow (hundreds of ms) to load.

3. Defered rendering has been enabled on DataTables: it avoid loading hundreds of mon images when only 10 rows are visible by default.

4. Images have a height to prevent row size from changing after the image is loaded.

TL;DR: 1s < 8s

### DIFF
--- a/mapadroid/patcher/__init__.py
+++ b/mapadroid/patcher/__init__.py
@@ -40,6 +40,7 @@ MAD_UPDATES = OrderedDict([
     (33, 'routecalc_rename'),
     (34, 'trs_stats_detect_raw_split'),
     (35, 'madrom_autoconfig'),
+    (36, 'pokemon_iv_index')
 ])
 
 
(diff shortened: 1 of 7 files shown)


ChaosEmpire:
would it be doable, to add an option for last hundos, like we have in shiny stats page
default one day, but selectable as time frame, 
that way it does not need to load much per default AND you would be able to see the hundos found between month xx and month yy ago

boby360:
I'm no developer, although wouldn't it be desirable to have all the pdfmake and other CDN JavaScript files on the same version across all of MAD?

MrJul:
> I'm no developer, although wouldn't it be desirable to have all the pdfmake and other CDN JavaScript files on the same version across all of MAD?

Yes it would! (Ideally those dependencies should be referenced only once in a central place.)
I just updated those libs in the other stats pages.

PR #986, opened 2020-09-03
Some UI fixes for MADmin (mainly for mobile)

georgeherby:
- Sorted the popup loading dialogue and moved to a central component
image

- Added some priority to columns for small screens
178 - iPhone X - 2020-09-03 at 1 55 07 PM

- Add expand/collapse for the responsive tables

Kapture 2020-09-03 at 13 52 44


cyphercris:
Loading gif is not shown on settings page. Format od adres should worki with domain.com, but probably not worki with domain.com/madminpanel/

PR #1010, opened 2020-09-21
plugins: handle exceptions in walk_package, ignore hidden directories

crhbetz:
Fixing two issues:

* previously, unhandled exceptions caused by faulty plugins would stop loading subsequent plugins, now it'll be visible which plugin caused issues and why, and we'll be able to continue with subsequent plugins

* recursing into hidden directories caused an exception because the contruction of package names as `package + '.' + child_pkg` with child_pkg starting with a dot seems to have caused invalid package names. There should be no reason to recurse into directories that are hidden / starting with a dot.

Found, applied and tested somewhat extensively during development of my recent poracle pvp plugin.
Running several other private and public plugins with these changes applied without problems.

### DIFF
--- a/mapadroid/utils/pluginBase.py
+++ b/mapadroid/utils/pluginBase.py
@@ -132,9 +132,9 @@ class PluginCollection(object):
         zipobj = zipfile.ZipFile(plugin_file_temp, 'w', zipfile.ZIP_DEFLATED)
         rootlen = len(folder) + 1
         for base, _, files in os.walk(folder):
-            if "__pycache__" not in base:
+            if "__pycache__" not in base and "/." not in base:
                 for file_to_zip in files:
-                    if file_to_zip != "plugin.ini":
+                    if file_to_zip != "plugin.ini" and not file_to_zip.startswith("."):
                         fn = os.path.join(base, file_to_zip)
                         zipobj.write(fn, fn[rootlen:])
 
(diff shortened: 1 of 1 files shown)


crhbetz:
This Travis failure is kinda ridiculous if you ask me 😄 

https://www.flake8rules.com/rules/W503.html
https://stackoverflow.com/a/57074416

fosJoddie:
I've added an ignore rule for W503 - that's the wackiest rule description I've seen in a while. 

I would love for someone to test this and just confirm it breaks nothing, but I cannot really see what it should break from looking at it on GH. I'll try and get it tested before too long

crhbetz:
Added one more little commit to exclude hidden stuff from packaging to `.mp` file.

My reasoning - hidden stuff in a published package doesn't make sense to me. It's usually hidden for a good reason. Like a `.git` folder, for example :trollface: 

It would break if your whole MAD installation is in a hidden folder somewhere in the directory structure, but again .. why? I'd embrace a more robust approach though.

Expl0dingBanana:
Has this PR been tested?

crhbetz:
I'm running it on my prod systems since before I started the PR without issues 🤷‍♂️ 

I'd like approval from @cecpk because of the modification to .mp file inclusion: `exclude hidden files/folders from plugin zipping` + general feedback regarding the hidden folder "issue" as detailed in my previous comment.

Expl0dingBanana:
@cecpk ping

PR #1012, opened 2020-09-22
PTC / Data Manager fixes

Expl0dingBanana:
* Fixed an issue where conversion would not apply the correct instance_id to ptc accounts being converted
 * Implemented additional search functionality within the data manager (.eq and .like)
 * PogoAuth accounts are now pulled correctly with the search functionality update
 * Updated requirements.txt to not use greenlet 0.4.17 as it caused issues with py3.7


Expl0dingBanana:
Received confirmation from asdfghjklyx that this resolved their problem.

PR #1020, opened 2020-10-07
Make it possible to show cell updates that are newer than a setting

fosJoddie:
I pushed the rest of these code changes in by accident (and nobody seems to notice).

This PR adds a new setting in madmin map that allows users to only show cell updates that are newer than a number of milliseconds (I want to change this to minutes or hours before mergin this PR)

### DIFF
--- a/static/madmin/templates/map.html
+++ b/static/madmin/templates/map.html
@@ -174,6 +174,12 @@ var setlng = {{ setlng }};
                 <input id="questcoordinateradius" type="number" v-model="settings.routes.coordinateRadius.quests" max="1100" class="form-control form-control-sm" />
               </div>
             </li>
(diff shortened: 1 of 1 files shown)


sn0opy (approved):

sn0opy:
Works just fine, but I think either the sidebar option needs a proper explanation or the default value should be changed to something like 48h. 

It wasn't 100% clear to me what kind of changes I should expect on the map when chaning the value in the input. Changing the value in the input box doesn't trigger any changes on the map either. Not sure if we should just trigger a redraw of all the cells here.

fosJoddie:
Needs better wording for sure, maybe even a little [?]  with tooltip.

Also should redraw on change, I didn't add it cause the whole madmin map js code is 🍝  and hard to follow

PR #1029, opened 2020-10-14
Remove repeated trashcan click in _clear_quests

crhbetz:
In earlier game versions, the confirmation dialogue to delete a quest wouldn't reliably pop up. Repeatedly clicking the trashcan symbol helped making sure we got the dialogue up.

Unfortunately, in recent game versions, clicking outside of the dialogue when it is already up causes it to close again. The repeated trashcan click is outside of the dialogue, thus it will close it and prevent confirming the deletion.

I observed this on a scanner device and confirmed it by manually trying it on my phone. Applying this fix improved the general quest scanning flow.


crhbetz:
Closing this PR as it's now obsolete with #1051 and the corresponding PD changes.

Issue #869, opened 2020-05-22
Check for finished quests


dkmur:
option in MADmin to check all devices on finished quests i.e. by taking screenshot of field research tab.

Expl0dingBanana:
I believe this is out of the scope of MAD as this sounds related to special research.  Is this the screenshot you were looking for?

JabLuszko:
If this is your use case then it would be probably much easier to handle that on MAD logs level (I mean atm you can probably grep for it) - if MAD needs to clear quests after 2 spins (or 1) few times in a row then we can assume that there is done quest waiting for you -> log something very specific / send alert.

dkmur:
That would work as well I guess, yet not something within the capabilities of this idiot 😇

muckelba:
I mean a single MADmin function for "open up the quest menu and take a screenshot" should be enough, right?

dkmur:
That was my thought, ideally to be able select a selection of devices in case you see drop in quest scan in an area

pki0:
Small suggestion:

If a big orange field ist detected during trashcan-scan (maybe only scanned if MAD found only 2 trashcans 3 times in a row) put a small yellow `!` or warning symbol on the status page in the row of the device.

dkmur:
Sounds good, whatever is the easiest way to implement I guess

spammer23:
But keep in mind - finished quests in stack will always show the orange field

PR #1030, opened 2020-10-14
Feature: Clear finished quests

crhbetz:
# Review and testing required !!!

Related issue: #869

The recent event bringing a "spin 3 pokestops" quest highlighted the problem of handling finished quests during quest scan. While it's not usually a problem on pure scanner accounts, events and changes can severely impact scan performance by introducing quests a quest scanner unintentionally finishes before it's able to delete them.

**The most stable way to handle this issue is usually to set `cleanup_every_spin` in area settings to `True`** (or using quest enhancer addon). However, this negatively impacts scan performance in dense areas, because the time and resource heavy ocr routines have to run more frequently.

However, this PR attempts to handle finished quests as far as a scanning software is able to. It works as follows:

1. If during quest deletion, less than the expected amount of trashcans is found, a check for finished quests will follow.
1. Using OCR, finished quests are identified by keyword
1. Finished quests are categorized according to screen position and color tint (stack, retrieveable, blocked have slightly different tints of orange)
1. Quests are handled accordingly - finished quests will be retrieved; when quests are blocked, it will be attempted to claim the weekly breakthrough and run. This will enable another 7 days of claiming quests
1. If quests remain blocked, an error will be logged saying the breakthrough has to be caught

In the end, this PR results in - ideally - requiring manual intervention once every 14 days to retrieve weekly breakthroughs, instead of possibly every day to clear up individual finished quests. It'll also explicitly identify the issue in the logs instead of the workers failing without an explicit cause.

Notes:

- the change in ` mapadroid/ocr/matching_trash.py` was necessary because quests blocked by the breakthrough were identified as having a trashcan - raising the threshold to `0.65` reliably fixed this without causing any false negatives
- successfully tested on a setup of >30 x96mini devices scanning > 2k quests daily. I achieved close to normal scanning performance despite the ongoing event with `cleanup_every_spin` set to `False`
- ...with default settings regarding screenshot filetype, quality
- I don't think it is necessary to make this feature optional as I can't see any negative impact - I'm open to other opinions on that, though
- **I consider further testing across different devices necessary**

### DIFF
--- a/mapadroid/ocr/matching_trash.py
+++ b/mapadroid/ocr/matching_trash.py
@@ -64,7 +64,7 @@ def trash_image_matching(origin, screen_img, full_screen):
 
         last_y_coord = 0
         res = cv2.matchTemplate(screen, resized, cv2.TM_CCOEFF_NORMED)
-        threshold = 0.5
+        threshold = 0.65
         loc = np.where(res >= threshold)
         boxcount = 0
         for pt in zip(*loc[::-1]):
--- a/mapadroid/ocr/pogoWindows.py
+++ b/mapadroid/ocr/pogoWindows.py
@@ -616,7 +616,7 @@ class PogoWindows:
 
     def __check_finished_quest_internal(self, image, identifier):
         origin_logger = get_origin_logger(logger, origin=identifier)
-        results = dict.fromkeys(["breakthrough", "finished", "blocked"], [])
+        results = {"breakthrough": [], "finished": [], "blocked": []}
         globaldict: Optional[dict] = {}
         diff: int = 1
         origin_logger.debug("__check_finished_quest_interal")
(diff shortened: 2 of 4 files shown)


spammer23:
Im working fine with cleanup_every_spin set to True - But sometimes the quest got solved anyway.
Will your changes work as backup in this case and clean the solved quest?

JimBim89:
> Im working fine with cleanup_every_spin set to True - But sometimes the quest got solved anyway.
> Will your changes work as backup in this case and clean the solved quest?

No this wont clean solved quests, but reduce the cases where quests are solved.

crhbetz:
It will eventually. To run efficiently, the check for finding finished quests only happens if less trash cans than expected were found - that's less than one in the case of running cleanup every spin.

This means that it'll only clean quests for you if you have three (or less if a trash can can't be found for another reason).
It should then clean them all in one go.

crhbetz:
> > Im working fine with cleanup_every_spin set to True - But sometimes the quest got solved anyway.
> > Will your changes work as backup in this case and clean the solved quest?
> 
> No this wont clean solved quests, but reduce the cases where quests are solved.

This is all about clearing finished quests (to the stack).

JimBim89:
Yeah I thought this was related to PR1029, will also test this ;)

crhbetz:
This is now obviously conflicting with the new AR quest mess, specifically the changes coming with #1051.

I will have to reconsider this whole feature and check if the upcoming PogoDroid will even allow to do this. So put on hold for now.

crhbetz:
Reworked to check for finished quests on the initial startup, and when the quest log is unexpectedly full (`Failed getting quest but got items`).

Working on recent master with updated PD keeping an AR quest in inventory. Testing would now be welcome :-)

Grennith:
Uhm this was left for ages I guess.... What's the state of the PR here?

crhbetz:
@Grennith been running this myself ever since I created the PR. I guess nobody cares anymore because we didn't get another "spin 3 pokestops" quest since then. It would still be relevant if this quest ever came back.

PR #1034, opened 2020-10-16
Count up and restart on repeated spinnable check failures

crhbetz:
Lately there were several scattered reports of devices behaving weirdly and getting stuck in a constant loop of the loglines `Can't spin stop - no map info in GMO!` and/or `Can't spin stop - couldn't find it closeby!`.

The devices would be stuck trying to click or spin stops and receiving a pink `Try again later` warning, or walking from stop to stop without interacting with them.

This indicates some sort of obscure issue(s) retrieving GMO data correctly containing the desired PokeStop. As also the game itself indicates an issue with the pink warning popping up, it seems a restart of the game app would be in place if this behaviour is shown persistently.

The number of 10 consecutive failed checks to trigger a restart was chosen arbitrarily. 🤷‍♂️

### DIFF
--- a/mapadroid/worker/WorkerQuests.py
+++ b/mapadroid/worker/WorkerQuests.py
@@ -153,8 +153,8 @@ class WorkerQuests(MITMBase):
         self.logger.debug("Getting time")
         speed = routemanager_settings.get("speed", 0)
         max_distance = routemanager_settings.get("max_distance", None)
-        if (speed == 0 or (max_distance and 0 < max_distance < distance) or
-                (self.last_location.lat == 0.0 and self.last_location.lng == 0.0)):
+        if (speed == 0 or (max_distance and 0 < max_distance < distance)
+                or (self.last_location.lat == 0.0 and self.last_location.lng == 0.0)):
             self.logger.debug("main: Teleporting...")
             self._transporttype = 0
             self._communicator.set_location(
@@ -786,8 +786,7 @@ class WorkerQuests(MITMBase):
                         self.clear_thread_task = ClearThreadTasks.QUEST
                     break
 
-            elif (data_received == FortSearchResultTypes.TIME or data_received ==
-                  FortSearchResultTypes.OUT_OF_RANGE):
+            elif (data_received == FortSearchResultTypes.TIME or data_received == FortSearchResultTypes.OUT_OF_RANGE):
                 self.logger.warning('Softban - return to main screen and open again...')
                 on_main_menu = self._check_pogo_main_screen(10, False)
                 if not on_main_menu:
--- a/mapadroid/worker/WorkerQuests.py
+++ b/mapadroid/worker/WorkerQuests.py
@@ -611,6 +611,9 @@ class WorkerQuests(MITMBase):
 
         gmo_cells: list = latest.get(PROTO_NUMBER_FOR_GMO).get("values", {}).get("payload", {}).get("cells",
                                                                                                     None)
+        pdlocation = latest.get("location", Location(0, 0))
+        pdlocation = Location(0, 0) if not pdlocation else pdlocation
+
         if gmo_cells == list():
             self.logger.warning("Can't spin stop - no map info in GMO!")
             self._spinnable_data_failure()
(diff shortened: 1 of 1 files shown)


Tabbomat:
I've been on this branch now for a couple of days, yesterday `self._spinnable_data_failure()` triggered a restart for the first time.
Log: https://pastebin.com/qDNzjjka - see line 303.
It seems to me that `self._spinnable_data_failcount` was not reset properly to 0 earlier.

On line 265, x96Mini1 started going to a new stop. While it couldn't find the stop closeby at first (line 274), it received the quest properly on line 290. I think that at this point `self._spinnable_data_failcount` should have been set to 0.

x96Mini1 starts going to the next stop. Again, it could not find it closeby (line 302), but immediately after that pogo gets restarted.

There are 11 occurrences of `Can't spin stop - couldn't find it closeby!` before the restart on line 303, therefore `self._spinnable_data_failcount = 0` must have happened at least once before line 303.

As far as I understand, the goal of this PR was to restart pogo after 10 consecutive fails for a single stop, not 10 fails alltogether.

If you need more logs from that night or if I censored the logs too much, please message me on discord.

crhbetz:
Thanks for testing!

I see where the counter fails to reset, however I'm not sure about how we can get into this state. There's something not going right with checking the distance to the pokestop, I assume.

I'll have to investigate further to make a correct fix.

mapadroid/worker/WorkerQuests.py
```diff
@@ -635,12 +638,14 @@ def _current_position_has_spinnable_stop(self, timestamp: float):
                 if fort_type == 0:
                     self._db_wrapper.delete_stop(latitude, longitude)
                     self.logger.warning("Tried to open a stop but found a gym instead!")
+                    self._spinnable_data_failcount = 0
```
fosJoddie, line 641: This might be the culprit.

We are looping over all pokestops and gyms we found in the GMO, but if we find a gym close by, we try and delete the pokestop at that location, and claim we tried to open a gym.

Actually I'm curious why this check is even here. This is the "are we at a place where there's a stop in the gmo" - not "we have tried to click a stop" ... anyhow. if gmo finds a gym here, it will fail spinnable check and reset spinnable

JabLuszko: But this is after the coords check and max accepted diff there is 0.00003. That gym would have been almost on top of pokestop, like ~4 meters MAX apart.

jfberry: I haven't followed this. But in my locality we have two gyms where people have moved pokestops right on top. Literally they show as underneath.  So these situations do exist.  These stops fail quest scan, but if you msg me on discord I will send you long,lat so you can see

crhbetz: That log line is misleading. We're not yet trying to open anything there. I want to reset the counter every time we get a "defined" result. So finding a gym where we expected a stop is a defined result. The stop being disabled/closed/on cooldown is a defined result.

On the other hand - not finding the stop at all or having "no map data" is not a defined result, because it could possibly be some type of hiccup.

crhbetz:
The issue is that when `_wait_for_data` times out, the following if clause falls through and the function goes on as if the spinnable check succeeded.

https://github.com/Map-A-Droid/MAD/blob/880958736d3fd5ec6087c10e1b223c386cba8681/mapadroid/worker/WorkerQuests.py#L677-L689

Thus what happens in @Tabbomat 's log is: The spinnable check fails once and adds to the counter, then a timeout happens in the quoted code, we proceed to (successfully) try to open the stop and get a new quest _without ever re-checking the spinnable status,_ and thus the timer doesn't get reset.

crhbetz:
I reworked this a bit to 

- base the location confirmation check on the actual location PD reports back to MAD
- make the spinnable check code more clear
- make it loop more often, because PD 1.8.1 apparently being able to get GMO more frequently caused more failed checks
- make it actually use the timestamp that's being passed to evaluate whether to wait for another GMO before the check

@Tabbomat I'd appreciate you giving this another test run. I ran it overnight and it's looking very good and stable for me.

PR #1039, opened 2020-10-19
Timezones fix

JabLuszko:
Raids were not showing on madmin maps if local timezone was before GMT/UTC (-), was working when it was after, +.
@Expl0dingBanana can you check if this thing also works on your strange UTC everywhere setup?


Expl0dingBanana:
Confirmed working on my instance

PR #1040, opened 2020-10-20
make questgen go fast

fosJoddie:
Only read quest related JSON once
Don't parse tasks as regex when we already have the condition objects at hand, or when regex is just a roundabout way to do str.contains()

### DIFF
--- a/mapadroid/utils/questGen.py
+++ b/mapadroid/utils/questGen.py
@@ -108,7 +108,9 @@ def questtype(quest_type):
 
 
 def rewarditem(itemid):
-    return items[str(itemid)]['name']
+    if str(itemid) in items:
+        return items[str(itemid)]['name']
+    return "Item " + str(itemid)
 
 
 def pokemonname(id):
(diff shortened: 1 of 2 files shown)


sn0opy:
needs a rebase

Expl0dingBanana:
To improve performance a bit more we can use try/except instead of "if key in list/dict". This is due to the key being checked twice (once during the in statement and once during the lookup).


```
import random
import time

MAX_RUNS = 5000000
MAX_SIZE = 100
ATTEMPTS = 5

lookups = []
for x in range(MAX_RUNS):
    lookups.append(random.randint(1,MAX_SIZE))

actual = []
for x in range(MAX_SIZE * 2):
    actual.append(x)

run_times = {
    "in": [],
    "try": []
}

for x in range(ATTEMPTS):
    start_time = time.time()
    for key in lookups:
        try:
            b = actual[key]
        except KeyError:
            pass
    run_times["try"].append(time.time() - start_time)

for x in range(ATTEMPTS):
    start_time = time.time()
    for key in lookups:
        if key in actual:
            b = actual[key]
        else:
            pass
    run_times["in"].append(time.time() - start_time)



print(sum(run_times["in"]) / len(run_times["in"]))
print(sum(run_times["try"]) / len(run_times["try"]))

>>> 1.70337743759
>>> 0.202215003967
```

fosJoddie:
superseeded by other PR

PR #1041, opened 2020-10-21
Add Quest Type/Template for New AR Task

Seandals:
This probably isn't the ideal implementation as it would be nice to know where the quest is going to send you but from my limited programming knowledge it doesn't seem to be saved into the database at this time.

At least it is better than "Unknown quest type placeholder: a" :)


fosJoddie:
Thanks for this, I only saw it after I already pushed my changes. I went for the solution "scan pokestop" since the target is always the pokestop the quest is at

PR #1042, opened 2020-10-21
Add Stickers to not deletable items

JabLuszko:
You can now click Stickers in inventory and it will open page full of stickers.
There is no trashcan on those stickers in main menu (for now, I can imagine Niantic QoL update).

Can't really get names in French/German, not translated in PokeMiners assets, will update later (or someone can comment).


JabLuszko:
Was told it's "Stickers" in German too, so I would assume it's still not translated and we need to wait for proper translation before merging this thing.
Thanks Chaotix

JabLuszko:
Okey, it's `Stickers` in French and English. It's `Sticker` in German.
Niantic re-inventing words I guess. So this is 'done' for now, lol.

PR #1044, opened 2020-10-22
Update the Dockerfile for new defaults

muckelba:
untested, but should work

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ RUN printf "deb http://httpredir.debian.org/debian stretch-backports main non-fr
 COPY . /usr/src/app/
 
 # Set Entrypoint with hard-coded options
-ENTRYPOINT ["python3","start.py", "-wm", "-os"]
+ENTRYPOINT ["python3","start.py"]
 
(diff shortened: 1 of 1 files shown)


memcoreDE:
It's, not working.
Since the update, i have an error.

`start.py: error: argument -pwhn/--pokemon_webhook_nonivs: ignored explicit argument 'true'`

sn0opy:
@memcoreDE you most certainly broke something in your setup then. `pokemon_webhook_nonivs` does still exist and you simply have to write `pokemon_webhook_nonivs` into your config file. It does not have any value.

memcoreDE:
Thanks.
But i have not changed the config. I have only updated the latest Commits/Docker-Image since 24h.
Ther is a issue with the configfile, i have disabled webhooks and have another errors.
`start.py: error: unrecognized arguments: --with_madmin=true --geofence_file_path=configs/geofences`  

Thanks, i will check the whole config.

sn0opy:
Read #announcements in Discord or https://github.com/Map-A-Droid/MAD/pull/1023

PR #1049, opened 2020-10-25
[RFC] Allow quest routes to always be recalculated dynamically

fosJoddie:
Adds a new setting to every quest area: "clear_route_every_time" - when set, this setting will make MAD always recalculate the route of a quest area instead of using the saved area. This will allow for faster questing in areas where stops tends to convert into gyms at a high rate, which will make MAD spend time trying to spin stops.

This commit also removes quite a bit of old unused code that looked very important, but was never used.

WARNING: testing this PR WILL add a new column to one table. If you can't figure out how to revert this or what table is affected on your own, DO NOT TEST!

I'm also looking for a code review to make sure I didn't add any code that doesn't belong in here by mistake

### DIFF
--- a/mapadroid/data_manager/modules/area_pokestops.py
+++ b/mapadroid/data_manager/modules/area_pokestops.py
@@ -113,6 +113,15 @@ class AreaPokestops(Area):
                     "description": "Cleanup quest inventory after every stop (Default: False)",
                     "expected": bool
                 }
+            },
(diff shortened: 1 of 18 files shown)


fosJoddie:
After running this for a while and thinking it over, I'm not sure this is the right way to go about it. A refactored worker manager is probably way more efficient in the long run.

PR #1058, opened 2020-11-10
Log streamer

Expl0dingBanana:
Implements a new page that allows for logs to be streamed to the browser and a download button for the current logfile.

### DIFF
--- a/mapadroid/madmin/routes/path.py
+++ b/mapadroid/madmin/routes/path.py
@@ -4,10 +4,12 @@ import os
 from mapadroid.madmin.functions import auth_required, get_quest_areas
 from mapadroid.utils import MappingManager
 from mapadroid.utils.functions import generate_path
-from mapadroid.utils.logging import get_logger, LoggerEnums, get_log_file
+from mapadroid.utils.logging import get_logger, LoggerEnums, get_log_file, AjaxSink
 
 
 logger = get_logger(LoggerEnums.madmin)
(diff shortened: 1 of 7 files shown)


Expl0dingBanana:
image

muddpunk:
I don't find I need to use logs too often, but I really like the idea of this for when I do. This would make it much easier for me personally. I like the idea of not having to go to my PC every time something feels off.

JabLuszko:
Works here, but can't see anything in Darkmode!!111!1111

gcm-explo1t:
1. You could replace 

while True:

in log_stream with some time dependent check, that it only runs for like 5-10 mins and then needs a refresh to solve JabLuszko's mentioned problem.

2. This method requires a log file, but why is this required, wouldn't it be possible to directly attach to the logger class and read out from there? The reason for this concern is, that some people (like me) don't actually write a log file to the system, as it creates gigabytes of log in a few minutes and pushes the discIO.

But in summary, nice Idea :)

C64Axel:
Nice Idea.
I use logstation for viewing all my logs in the Browser. Maybe you can add such thinks like "Pause" and "Regex"?
Regex is very usefull if you only want to see logs from one Origin or for example only loglines from the updater.

Grennith:
Given the conflicts, was this PR ready to be merged back then? If so, mind fixing conflicts?

PR #1063, opened 2020-11-29
Implement caching for incoming data before DB insert/update

sn0opy:
This PR implements basic caching via Redis for data sent by devices. Data like stops, gyms, mons etc. are cached by certain identifiers. Some will auto-invalidate the data once changed, others will wait for a certain timeout to remove keys from the cache.

It requires the `redis` dependency to be installed via pip and a running Redis server. Redis host + port can be configured optionally. See example config on how to enable caching.

Certain things are being removed from the cache after 15 minutes and will then be re-inserted/updated in the DB + added back to the cache afterwards. This is somewhat of a precaution for now to make sure things on the map still get update to see if everything's working correctly. Cells aren't cached at all to see a permanent movement of workers on the map.

Spawnpoints could probably added to the cache as well.

| Type | Method |
| ---      | ---       |
| Mons | Cached by encounter_id. Auto-invalidates after despawn time | 
| IVs | Cached by encounter_id + weather_boosted value. Should invalidate if the weather changes. Auto-invalidetes after despawn time |
| Spawns | Not cached |
| Stops | Cached by stop id and last_modified timestamp. Should invalidate if new incidents (Team Rocket) appear |
| Stop details | See line above |
| Quests | Not cached |
| Gyms | Cached by gym ID + last_modified timestamps. Should invalidate if gym defenders or raid changes. Auto-invalidates after 15min |
| Raids | Cached by gym ID + raid boss + raid end time. Auto-invalidates after 15min | 
| Weather | Cached by cell id + weather values. Auto-invalidates after 15 min |
| Cells | Not cached |


mapadroid/cache/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+
+logger = get_logger(LoggerEnums.system)
+
+
+def get_cache(args):
+    cache = NoopCache()
+    if args.enable_cache:
+        try:
+            import redis
+            cache = redis.Redis(host=args.cache_host, port=args.cache_port, db=args.cache_database)
```
Expl0dingBanana, line 12: We should try to keep this connection open to reduce the overhead of the connection process. Maybe if we initialize it we should store it as a singleton to save some of the processing time of re-opening and pining.

mapadroid/cache/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+from mapadroid.utils.logging import get_logger, LoggerEnums
+
+logger = get_logger(LoggerEnums.system)
+
+
+def get_cache(args):
+    cache = NoopCache()
+    if args.enable_cache:
+        try:
+            import redis
```
Expl0dingBanana, line 11: This import should happen during file initialization with the try/except. This will reduce (a small) amount of overhead of re-importing. We can switch ImportError to NameError within the function to keep the same functionality

PR #1065, opened 2020-12-06
Development Workflow Improvements

Expl0dingBanana:
This PR allows for easier creation of a clean MAD environment with a clean database through `make`. It also implements pre-commit which allows for commands to be run prior to commiting to help ensure code quality. As of now this only works on *nix systems due to the current commands being run in Makefile.

How to use (note there are more but these will be the common ones):
 * `make setup`: Install pre-commit and git message template (this one should be run to ensure basic code quality is enforced)
 * `make clean`: Clean the tox environments to be re-created
 * `make build`: Build the container. This only needs to be re-built it any dependencies change
 * `make unittests`: Run python3.7 tox build
 * `make tests`: Run all tox environments 
 * `make up` / `make down`:  Startup or Shutdown the MAD and MariaDB containers
 * `make shell` / `make root-shell`: Access to MAD container as dockeruser or root
 * `make shell-py3{6,7,8}`: Access to MAD container with the specified python version

The build environment uses two containers, a pre-built MariaDB container, and a on-the-fly built Map'A'Droid container. The Dockerfile has been updated to include a second target that installs all required dependencies for testing against multiple versions of python. The multiple versions are handled through pyenv and is installed during container building. The docker-compose file has been updated to read from a file to ensure the correct user and passwords are set for MAD to communicate with the database.

The following pre-commit plugins have been added to the commit pipeline:
 * gitlint: Checks against the git commit message. Config is stored in .gitlint
 * flake8
 * isort: Sort imports automatically
Hooks:
 * trailing-whitespace: Remove any trailing whitespace at the end of lines
 * end-of-file-fixer: Ensure there is a newline at the end of the file. If not, it will be automatically added

Tox is being used as an envrionment manager to ensure we test against all required versions of python and execute the correct steps. The previous .flake8 file has been removed as its contents have been added to tox.ini This PR implements the following tox test steps:
 * pytest tests/ (Execute all required tests in the tests/ folder)
 * flake8 (Validate we meet all flake8 requirements)


.dev.env
```diff
@@ -0,0 +1,9 @@
+MYSQL_ROOT_PASSWORD=password
+MYSQL_DATABASE=rocketdb
+MYSQL_USER=rocketdb
+MYSQL_PASSWORD=rocketdb
+
+THERAIDMAPPER_DBIP=mariadb
```
Grennith, line 6: Guess we could just use the configuration above for user and password?

Grennith (commented):
All in all, looking good

.dev.env
Expl0dingBanana: Each container is expecting different variable names which we have them defined twice. The DBIP is a bit different since that is the name of the MariaDB container that is used for DNS resolution

fosJoddie:
How is this supposed to work for people who code in a Windows environment?

I install gnumake and ran `make setup` which returned this:

> make setup
>process_begin: CreateProcess(NULL, which pip, ...) failed.
>Makefile:12: pipe: No error
>process_begin: CreateProcess(NULL, which pre-commit, ...) failed.
>Makefile:19: pipe: No error
>Makefile:20: ***     Install pre-commit @ https://pre-commit.com/#install then run
>    source ~/.profile.  Stop.

I'm not sure what to make of this error tbh

PR #1067, opened 2020-12-15
encounter_all setting for *_mitm areas

crhbetz:
This PR adds the `encounter_all` setting to the three *_mitm areas. It will default to `False`, so the scanning behaviour will not change until intentionally set.

When enabled, the Worker will stay at the same location until it can be assumed that all Pokemon at that location have been encountered ("IV-scanned"). To achieve this, we compare the encounter_ids known to be already encountered to those present in the newly fetched data. We count all unique Pokemon IDs (unique species) that haven't been encountered yet and divide this number by 3 to know how often we need to wait. This is based on the knowledge that PogoDroid on the device will encounter 3 unique species per GMO data received. After waiting this number of times, the Worker will continue as usual.

The intention of this additional option is to get the most encounter data out of every location change. Through the additional waits, the device will handle less locations per time, but achieve a near 100% encounter rate for the Mons it sees. Optimizing for locations per time or encounter rate is personal preference, thus I designed it as an optional feature that's disabled by default.

The logic has been tested successfully for some weeks already. Review and testing is required for this pretty and database changing version :-)


ChaosEmpire:
Great PR, 
98% or more IV Rate per area now

mag3sticl:
Using this PR and it works great. Like above, it pushes my IV% up to 98%.

FimoJones:
Working well. 99% of my spawns are now being covered as a result of this PR.

PR #1068, opened 2020-12-17
Add is_ar_scan_eligible flag to pokestop and gyms

JabLuszko:
This is totally useless now, nothing supports this, but could be useful in future.

gyms & quests webhook uses that flag. We don't have webhook for pokestop changes (and that would need a lot of logic), so I ended up and put that in quests - once day, should be good enough.

### DIFF
--- a/mapadroid/patcher/add_is_ar_scan_eligible.py
+++ b/mapadroid/patcher/add_is_ar_scan_eligible.py
@@ -11,7 +11,8 @@ class Patch(PatchBase):
             ADD COLUMN `is_ar_scan_eligible` tinyint(1) NOT NULL DEFAULT '0';
         """
         try:
-            self._db.execute(alter_pokestop, commit=True, raise_exec=True)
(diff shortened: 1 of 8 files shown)


mapadroid/patcher/add_is_ar_scan_eligible.py
```diff
@@ -0,0 +1,27 @@
+from ._patch_base import PatchBase
+
+
+class Patch(PatchBase):
+    name = 'Add add_is_ar_scan_eligible flag to pokestop and gyms'
+    descr = 'New Protos, more useless data!'
+
+    def _execute(self):
+        alter_pokestop = """
```
Expl0dingBanana, line 9: We should check to see if the column exists prior to executing the sql. This allows the patch to be run multiple times without experience any issues. Here is a good example of that in-use

https://github.com/Map-A-Droid/MAD/blob/master/mapadroid/patcher/pokemon_iv_index.py#L18

JabLuszko:
There you go.

PR #1070, opened 2020-12-21
Move Dockerfile back to project root

DrMurx:
Many standard build tools expect the Dockerfile in the project root. It can be considered similar to a Makefile which is also in the project root.

Other supplementary files for a docker build or a docker development environment can of course stay in the `docker/` directory.


sn0opy:
Makes sense to me.

@DrMurx would you like to rebase this?

DrMurx:
> would you like to rebase this?

@sn0opy Done

PR #1071, opened 2020-12-21
Make one of the datatables rows different color in darkmode

JabLuszko:
Not a great color choice, but at least they stand out a little now. Still readable, just more ugly.


Expl0dingBanana:
Old CSS:
image


New CSS:
image

PR #1077, opened 2020-12-27
Restore delete button

Expl0dingBanana:
Fixed a regression where the delete button no longer appeared for uploaded APKs. This bug was created during #932 when the name changed from "jobType" to "JobType" 

Updated the file-browser to default to *.apk files as they are the only extensions accepted (.txt is accepted too but i don't see how that is beneficial here...)


Expl0dingBanana:
image

PR #1079, opened 2020-12-27
Fix a stats issue

Expl0dingBanana:
* Fixed a name collision that caused state pages not to load


Expl0dingBanana:
Validated working

https://discord.com/channels/465247740553592832/467318193141317659/792852276208992346

PR #1081, opened 2020-12-29
Wizard fixes

crhbetz:
1. Make the Wizard "Update Everything" do the filesize check for RGC/PD before downloading (this check is run when clicking the "Search for new version" button in the individual lines above).
2. Improve the pogo update logic to make the logging easier to understand and to enable it to download "intermediate" versions. Example: `195.0` is installed. `195.1` is available to be downloaded through gplay. `195.0`, `195.1` and `195.2` are supported via the `version_codes.json` file. Previously, downloading `195.1` in this situation was not possible while `195.2` was not yet available through the store, thus we would remain at `195.0`. This behaviour should be fixed.
3. Fix a wrong global variable reference (was `VERSIONCODES_GITHUB`, corrected to `VERSIONCODES_URL`)

@Expl0dingBanana - please review.

### DIFF
--- a/mapadroid/mad_apk/wizard.py
+++ b/mapadroid/mad_apk/wizard.py
@@ -311,7 +311,8 @@ class APKWizard(object):
                         version_code = current_version_code
                         version_str = current_version_string
                     else:
-                        return None
+                        version_code = latest_supported["versionCode"]
+                        version_str = latest_supported["version"]
                 elif current_version_code and store_version_code <= current_version_code:
                     logger.info(f"Latest store version is {store_version_string} while {current_version_string} is "
                                 "already installed. Unable to find a newer version")
--- a/mapadroid/mad_apk/wizard.py
+++ b/mapadroid/mad_apk/wizard.py
@@ -287,32 +287,40 @@ class APKWizard(object):
             if latest_pogo_version[architecture] is None:
                 return latest
             latest_supported = latest_pogo_version[architecture]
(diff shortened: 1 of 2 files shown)


crhbetz:
Looking into the failing tests ...

crhbetz:
Tests passed / adapted to new behaviour. Now good to go from my side.

mapadroid/mad_apk/wizard.py
```diff
@@ -280,32 +287,41 @@ def find_latest_pogo(self, architecture: APKArch) -> Optional[str]:
...
             latest_supported = latest_pogo_version[architecture]
-            current_version = self.storage.get_current_version(APKType.pogo, architecture)
-            if type(current_version) is not str:
-                current_version = None
+            current_version_string = self.storage.get_current_version(APKType.pogo, architecture)
+            current_version_code = None
+            if type(current_version_string) is not str:
+                current_version_string = None
+            if current_version_string:
+                ls: dict = {architecture: current_version_string}
```
Expl0dingBanana, line 295: We should use a more descriptive variable name than `ls`.

Expl0dingBanana:
Looks good to me. Thanks for these fixes!

mapadroid/mad_apk/wizard.py
crhbetz: Done. Had to add some sort of suffix because we already have a variable `latest_supported`.

PR #1084, opened 2020-12-31
IPv6 support for webhooks

loewe:
Using index instead of rindex finds the first closing bracket and not the one of the IPv6 adress.

### DIFF
--- a/mapadroid/webhook/webhookworker.py
+++ b/mapadroid/webhook/webhookworker.py
@@ -503,7 +503,7 @@ class WebhookWorker:
             url = webhook.strip()
 
             if url.startswith("["):
-                end_index = webhook.rindex("]")
+                end_index = webhook.index("]")
                 end_index += 1
                 sub_types = webhook[:end_index]
                 url = url[end_index:]


loewe:
IPv6 uses square brackets in URL's

f.e. http://[2001:0db8:85a3:08d3::0370:7344]:8080/

Webhooks can be prefixed with the type which leads to [raid]http://[2001:0db8:85a3:08d3::0370:7344]:8080/
The method tries to split and retrieves the type with rindex as "raid]http://[2001:0db8:85a3:08d3::0370:7344" where it
should only retrieve "raid" and the URL as ":8080/" which creates a urllib error.

Using index fixes the behavior. Was there another reason using rindex?

Expl0dingBanana:
I think the correct fix for this should be [regex](https://pythex.org/?regex=(%5C%5B%5CS%2B%5C%5D).*%3A%2F%2F&test_string=%5Braid%5D%5Bmon%5Dhttp%3A%2F%2F%5B2001%3A0db8%3A85a3%3A08d3%3A%3A0370%3A7344%5D%3A8080%2F&ignorecase=0&multiline=0&dotall=0&verbose=0) and https://docs.python.org/3/library/urllib.parse.html

loewe:
@Expl0dingBanana Thanks for your fast reply. regex or "urllib.parse" would also be useful to validate the URL, too. But that must be done after the splitting with "index". At earliest after line 509.

Expl0dingBanana:
What happens if you have [raid][mon]? Using index will skip the mon portion. and generate an invalid URL. The regex portion is a link to a regular expression that should solve that issue.

JabLuszko:
> What happens if you have [raid][mon]?

This is not supported. 

`[raidpokemon]` or `[raid pokemon]` or even `[raid:pokemon]` - anything except comma :)

loewe:
@Expl0dingBanana Could we please merge that PR? I need to rebase with every pull.

PR #1086, opened 2021-01-06
Fix a TypeError exception in __find_latest_head in wizard.py

crhbetz:
The issue was brought up on discord and is caused by changes from PR #1081.

A `TypeError` exception was thrown when trying to download RGC or PD through the wizard with no pre-existing version on the MAD instance. I could reproduce the problem locally and fix it with these changes.

It was not considered that the packages variable could be `None` when no version of the requested package is currently installed to
the MAD instance. This commit/PR will introduce code to handle this situation.

### DIFF
--- a/mapadroid/mad_apk/wizard.py
+++ b/mapadroid/mad_apk/wizard.py
@@ -248,15 +248,17 @@ class APKWizard(object):
         """
         update_available = False
         (packages, status) = lookup_package_info(self.storage, package)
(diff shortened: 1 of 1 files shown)


crhbetz:
Failing checks were caused by https://github.com/Map-A-Droid/MAD/commit/b097c780037796b12ed562207bb2cd831c7e51ae 🤷 

@cecpk 😘

PR #1089, opened 2021-01-12
Add script to truncate quests.

Artanicus:
Sometimes it's easier to run a script than some manual SQL.


Artanicus:
Anything you want tweaked or modified? Or is this considered a bad idea?

PR #1092, opened 2021-01-15
Pythonic locking

Grennith:
Changes in a nutshell:
- Locks (not Semaphores - which we have some still) are now used via `with` rather than try-finally
- "unhanded exception" -> "unhandled exception"
- WorkerQuests::_try_to_open_pokestop now returns an enum as we had a Tuple[bool, bool] before which imho would not properly cover the permutations of results possible
- MITMBase::_wait_for_data now cuts off the decimal places of the timestamp passed as PD uses integers (long) values thus sometimes resulting in MAD waiting for data that was received before the timestamp used by MAD internally (being float)

### DIFF
--- a/mapadroid/db/PooledQueryExecutor.py
+++ b/mapadroid/db/PooledQueryExecutor.py
@@ -39,11 +39,10 @@ class PooledQueryExecutor:
             "password": self.password,
             "database": self.database
         }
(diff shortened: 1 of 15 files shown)


Expl0dingBanana:
Looks good to me. The use of enums really improves the readability of the code!

PR #1097, opened 2021-01-19
Actually set cache on pokestops

JabLuszko:
15 minutes for now. Should probably use lure/incident end times?


sn0opy:
> 15 minutes for now. Should probably use lure/incident end times?

we do. The cache key contains the last_modified timestamp which changes once an incident or lure starts. May need changes since I'm unsure if the timestamp also changes once incident/lure ends. Things shouldn't break if it doesn't update once the lure/incidents ends though.

PR #1109, opened 2021-02-24
Implemented route editor. Fixes #1057

### LINKED ISSUE #1057: Visual route editor
Hi,
adding a route editor on the madmin panel could be a real improvement.
I mean, currently changing a stop point on a device route require manually inserting coordinates, a future improvement should be permit editing route points as from already happen on area.
Optimizing routes is very time consuming and I think that a area-like editor will dramatically improve the speed and the yield of the devices.

MrJul:
With this PR, `mon_mitm` routes can now be edited.

It's based on the geofence editor and works the same (click on a route, choose *edit* in the popup, drag/add/remove points, click *save* in the popup): common code for the two editors has been extracted.

Once edited, a route won't be active in memory until settings are applied. The main problem here is that refreshing the map without applying settings would get back the "old" applied/unedited route from memory. To avoid this, the map now shows an additional "unapplied" route. 

To avoid confusion, there are several special rules regarding this unapplied route:
 - It only exists if it's different from the applied route (= the route has been edited)
 - After saving, the edited/unapplied route is automatically made visible and the applied route is hidden, making it easy to continue editing the current route if needed.
 - When settings are applied or when MAD is restarted, the unapplied route ceases to exist and the normal route is shown instead (if the unapplied one was visible before).

This may seem a bit complicated while reading but from an user point of view it's mostly transparent. Unless comparing the applied and unapplied routes is needed, the user simply edits and saves a single route without having to toggle anything, as easy at is was with geofences.

To make things clearer in the sidebar, sub-routes and unapplied routes are now visually indented and tagged as such:
image

Other map changes:
 - Route lines are thicker to make them easier to click
 - The contrast between areas and routes has been improved (it's still not perfect but it should help with editing routes)

### DIFF
--- a/mapadroid/data_manager/modules/routecalc.py
+++ b/mapadroid/data_manager/modules/routecalc.py
@@ -129,21 +129,15 @@ class RouteCalc(Resource):
     def get_json_route(self, coords: List[Tuple[str, str]], max_radius: int, max_coords_within_radius: int,
                        in_memory: bool, num_processes: int = 1, algorithm: str = 'route', use_s2: bool = False,
                        s2_level: int = 15, route_name: str = 'Unknown') -> List[Dict[str, float]]:
+        if not in_memory:
+            saved_route = self.get_saved_json_route()
+            if len(saved_route) > 0:
+                logger.debug('Using routefile from DB')
+                return saved_route
+
         export_data = []
         if use_s2:
             logger.debug("Using S2 method for calculation with S2 level: {}", s2_level)
-        if not in_memory and \
-                (self._data['fields']['routefile'] is not None and len(
-                    self._data['fields']['routefile']) > 0):
-            logger.debug('Using routefile from DB')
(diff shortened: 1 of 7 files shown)


dkmur:
loaded it to test instance, played around with some routes.....great stuff. Merge please 😇

Expl0dingBanana:
There are a few flake8 fixes that need to be made. I havent reviewed the PR yet but looked into the build failures

$ python3 -m flake8
./mapadroid/madmin/routes/map.py:4:1: F401 'flask.redirect' imported but unused
./mapadroid/madmin/routes/map.py:4:1: F401 'flask.url_for' imported but unused
./mapadroid/madmin/routes/map.py:7:1: F401 'mapadroid.data_manager.DataManagerException' imported but unused
The command "python3 -m flake8" exited with 1.

MrJul:
Whoops, fixed those!

PR #1120, opened 2021-03-19
More Pokemon

ccev:
# Changes

## Config changes
- New arg `scan_nearby_mons` to enable nearby scanning (for both nearby_stop and nearby_cell types)
- New arg `scan_lured_mons` to enable lure pokemon (iv and noiv)
- Removed arg `pokemon_webhook_nonivs`
- Pokemon Webhooks can now be further defined based on the seen_type (look below for details). So you can use `[encounter nearby_stop]https://url.com` if you just want webhooks for IV mons or nearby_stop mons. `pokemon` is an alias for `encounter` to ensure backwards-compatibility
  - Spezus' pvp-plugin is just sending `encounter` and `lure_encounter` types. If you want more webhook-types, you have to configure them in your MAD config. If you're using the plugin, make sure to not put `encounter`, `lure_encounter` or `pokemon` in your MAD config

## New Pokemon-types

The game is sending more Pokemon than MAD is currently processing. With this PR, MAD can now properly handle them. 

There's also a new column in the `pokemon` table called `seen_type` that stores the type of the pokemon. Possible types:

- ### encounter

  - A regular wild Pokemon that got its IV scanned

- ### wild

  - A regular wild Pokemon without IV

- ### lure_wild
  - A Pokemon spawned by a lure without IV
  - The lured Pokestop's ID this Mon is at i saved in the `fort_id` column
  - Webhooks have `fort_id`, `fort_name` and `fort_url` fields.
  - `spawnpoint_id` is 0 and the lat/lon are taken from the lured Pokestop
  - Redis-cached based on encounter_id for 3 minutes

- ### lure_encounter
  - Lure encounter protos have like no data except the encounter values, so to have a proper lure_encounter, it requires a lure_wild version first
  - Redis-cached based on encounter_id for 3 minutes
  - Requires a PD version that isn't out yet (at the time of writing this)

- ### Nearbies
  - Each Level 15 cell has a limit of 6 nearby Pokemon
  - disappear_time is set to a 15-minute timer
  - Redis-cached based on encounter_id and mon_id for 1 hour. DB is only touched if there's no entry with that encounter_id already
  - It also factors in redis entries for wild and encounter versions of the Pokemon
  - ### neaby_stop
    - If there's a Pokestop in view-distance that has a Pokemon close to it, we receive Pokemon that are within 50m (?) around that Stop. No coords, spawnpoint_id or encounter data
    - The Pokestop's ID this Mon is at i saved in the `fort_id` column
    - Webhooks have `fort_id`, `fort_name` and `fort_url` fields.
    - `spawnpoint_id` is 0 and the lat/lon are taken from the Pokestop
  - ### nearby_cell
    - If there are Pokemon within 200m of your location and the L15 cell they're in has less than 6 nearby_stop, you'll receive nearby_cell mons
    - The cell's ID is saved in the `cell_id` column
    - Webhooks have `cell_id` and `cell_coords`
    - `spawnpoint_id` is 0 and the lat/lon are the middle of that cell

## Route/area changes

mon_mitm areas get a new setting: `nearby_cell_mode`. (For experts 🙂)
- Instead of waiting for wild_pokemon, a worker will wait for nearby_pokemon. 
- This uses a 200m range instead of 67m.
- If you enable init, nearby_cell_mode is internally overwritten with False
- If nearby_cell_mode is enabled, the `coords_spawns_known` is ignored

## New Stats table
- `trs_stats_detect_seen_type` acts as an extension to `pokemon`. It saves the times at which each pokemon was first seen in a certain way
- It's cleared by deleting all entries that aren't present in the `pokemon` table

## MADmin changes

- Pokemon Popups show the seen_type
- Cellupdate Popups show if that cell has seen nearby_cell Pokemon before or not
- Nearby Pokemon have randomly offset coordinates (Lure Pokemon keep its Pokestop coords)

## Misc changes
- Wild and encountered mons are now cached with the mon ID in case an event changes the species at that spawn

### lure_encounter
If you want to grab lure_encounters from the database, make sure to have your query look something like this:

```sql
SELECT * FROM pokemon WHERE seen_type = 'lure_encounter' AND latitude != 0;
```

this makes sure that your lure encounter has the lure_wild information in it.

### Database changes
- Add `fort_id`, `cell_id` and `seen_type` to the pokemon table.
    - `seen_type` specifies what kind of pokemon the entry is. An enum that can be one of the following: `wild`, `encounter`, `nearby_stop`, `nearby_cell`, `lure_wild `, `lure_encounter`
    - `fort_id` is used for nearby_stop, lure_wild and lure_encounter mons, `cell_id` for nearby_cell mons. For every other type, they're set to NULL
- Removed `encounter_id` from pokestop (this was a bug)
- Add `nearby_cell_mode` to seetings_area_mon_mitm

### Webhook changes
- Add `seen_type` to all pokemon webhooks. Look above for the possible values. Always sent as string.
- nearby_stop mons will have the following extra fields: `pokestop_id`, `pokestop_name`, `pokestop_url`. self-explainatory
- nearby_cell mons will have the following extra fields: `cell_id`, `cell_coords`. The latter being a list of the 4 edge-points of that cell.
- `encounter_id` in mon webhooks is now sent as string (poracle wanted this because node doesn't like big numbers)

### PD data
PD POST request examples for the different new types (so we're all on the same page here)

## nearby_stop
```js
{
  "id":29,
  "fort_id":"f28711eb128a8723f38c8d8e2823",
 "fort_url":"http:\/\/lh3.googleusercontent.com\/zugqduzkvDwjhfvbjkwhbe278gbJKWVHBjkbvJHVD2873762VKhvd",
  "encounter_id":71053162282263986536,
  "distance_to":498.6898193359375, // the distance between the account and stop
  "display":{
     "costume_value":0,
     "gender_value":2,
     "is_shiny":false,
     "form_value":776,
     "weather_boosted_value":0,
     "alignment":0,
     "pokemon_badge":0,
     "current_temp_evolution":0,
     "temporary_evolution_finish_ms":0,
     "temp_evolution_is_locked":false,
     "original_costume":0,
     "display_id":71053162282263986536
  }
}
```

## nearby_cell
```json
{
  "id":29,
  "fort_id":"",
 "fort_url":"",
  "encounter_id":71053162282263986537,
  "distance_to":0,
  "display":{
     "costume_value":0,
     "gender_value":2,
     "is_shiny":false,
     "form_value":776,
     "weather_boosted_value":0,
     "alignment":0,
     "pokemon_badge":0,
     "current_temp_evolution":0,
     "temporary_evolution_finish_ms":0,
     "temp_evolution_is_locked":false,
     "original_costume":0,
     "display_id":71053162282263986537
  }
}
```

## lure_wild

```json
{
"id": 1,
"latitude": 0,
"longitude": 0,
"spawn_id": "f28711eb128a8723f38c8d8e2823",
"encounter_id": 71053162282263986538,
"expiration_timestamp": 1618759767000,
"display": {
  "costume_value": 0,
  "gender_value": 0,
  "is_shiny": false,
  "form_value": 0,
  "weather_boosted_value": 0,
  "alignment": 0,
  "pokemon_badge": 0,
  "current_temp_evolution": 0,
  "temporary_evolution_finish_ms": 0,
  "temp_evolution_is_locked": false,
  "original_costume": 0,
  "display_id": 71053162282263986538
}
```

# Possible improvements
Some improvements that I thought may be useful for some people
- Space out nearby mons on a stop instead of setting them all to the same location
    - Update: I have decided against doing this, as I think the map frontend should handle it
- Make the default despawn time configurable (that value can then also be used for spawns with unknown endtimes)
- Grennith told me `"wenn man das noch sinnvoller machen will baut man ne prioQ dir korrekter arbeitet als die aktuelle bei iv_mitm und eiert einfach alles runter"` which I think means that iv_mitm could use a better prioQ algorithm to better work with nearby scanning
    - iv_mitm could generally use some improvement, especially for nearby_cell mons

## Nearby Spawn Matching
Nearby Spawn Matching means trying to predict the spawnpoint a nearby mon would be at, based on the data we have. I have played around with it and came to the conclusion, that it'd be best done in a plugin. I decided to not implement it for now, but below is how I imagine it'd work. (maybe someone else wants to do it or for future-me. DbPogoProtoSubmit already has a code snippet that kinda works)

It'd work the following:
- trs_spawn gets a new column `nearby_stop`. This is the StopId of the nearest stop that is within 50 meters of the spawnpoint
- When a nearby mon is found, search for spawns connected to this Stop
- Rule out stops that already have an active mon on it or are currently not active (and are within the current event's time)
- Then have 2 modes:
    1) If there's only one point, use that. Otherwise, skip it and just use the default values
    2) Assuming you fully cover the area and constantly see every stop, you could use the spawntime with the spawntime that is closest to right now

For this to fully work, you need accurate data. Handling spawnpoints is another story, but the plugin would have to do the following:
- Look for new spawnpoints to update with a nearby_stop
- Keep up with Pokestop changes: New Stops, Removed Stops, Stops converted to Gyms, Gyms converted to Stops, edited Stops. Especially hard would be taking care of Stop Edits

It'd be a nice feature to have, but really it's just taking a lot of effort and processing to tell you something you probably already know: The Mon is near the Stop and probably has a perfect DST.

# Usecases
To demonstrate how useful nearby scanning can be, here's some possible usecases:
- Have a big area but not enough funding? Use a raid_mitm worker (with stops enabled) and one (or multiple) iv_mitm worker with a limited mon list. This will allow maximum area with minimum devices. I think this is also how RDM does it.
- Set one of your devices to iv_mitm with a limited mon list (e.g. unown, axew, lake trio, gible, noibat) and leave your other devices running mon_mitm. Now, if a good Mon spawns next to a stop, your map will (in most cases) find that mon almost instantly. The iv_mitm worker will then go there and you get the correct despawn time, IVs and coords
    - You can also just set the iv_mitm are to the top of your walker with type `coords` and max device set to 1. So whenever something good spawns, one device checks the IV and then goes back to scanning normally
- Just update MAD and have nonIV webhooks enabled. Enjoy your super early rare notifications. Even if they have no exact location and despawn time - again, most of the time, they'll have a perfect timer and the exact location doesn't matter anyway.

## My personal setup

NOT: This it outdated and doesn't fully work as described. Basically, the "overflow" of devices doesn't really work. Also iv_mitm will never "run out of coords".

image

I have my walker like this. Each quest/mon area represents one town I'm scanning. The walkerarea at the top of the list is an iv_mitm area that covers everything with a few relevant super-rares in its mon list. There's also a raid area that I've always used to primarily scan raids.

With this config, whenever a rare nearby mon is found, some device leaves its area, TPs to that rare mon, scans it, then goes back to scanning. This ensures that all rare mons are scanned instantly while no device is standing around doing nothing.

Since my mon areas are constantly covering 100% of stops in the towns, when a rare mon appears near one of those stops, I get a notification as soon as it spawns in-game. Then, the iv_mitm worker goes there and checks it within the next 30-60 seconds, confirming that it is indeed a perfect DST and to check its IVs. This is very nice for Unown notifcations, but also to find good-IV Gible (or similiar). When everything is working normally, I usually receive the IV notifications with 27-28 minutes time left.

Since the raid area is meant to scan Raids, Mons are now just a byproduct there. Still, the PR gets me about 5000-6000 mons/hour in that area with 100 of them being IV scanned. Of course, most of them have average DSTs. In the month I've been using the PR, it already allowed me to scan 1 Unown there.

I'm also in the process of testing a nearby_cell_mode area to cover a huge rural area. But from my limited data/observations, I'd say that it's not too good.


sn0opy:
Since MAD is not actively using the nearby data (beside webhooks), I'd rather go with an opt-in setting instead of an opt-out one. Nearby mons are not shown on the map, nor does MAD use it for route calculations. So rather disable it and let people decide, if they really want it.

bbdoc:
What will happen if a nearby mon is found, sent to webhook, and then found again by a normal worker (including correct location, iv, timers). Webhook will consider it was already sent and not send it again ? So you end up with lots of mons having no info while they would normally have.

ccev:
It depends on how your Webhook Receiver handles it. I tested it with Poracle and it will re-send the Pokemon with its information after MAD has found it.

ChaosEmpire:
from my point of view after 1 day
Nearby are just regular spawns, I havent found 1 pokemon on map, which hasnt had a real spawnpoint when moving a worker there

that way it opens the possibility of better handling for PrioQ and IV List sorted based scanning

RocketMad is at the moment not able to dynamicly show the change from stop location to real location, it needs a reload

madmin map can handle this already

jfberry:
Confirm poracle is designed to handle the no-IV to IV situation (and change in IV if retransmitted eg after weather change) without interfering with de-duplication logic

ccev:
FYI for those who might want to test the PR: I've just added a feature where it tries to match a Nearby mon to a Spawnpoint, which is a relatively resource-intensive task. Tomorrow I'll try to put a proper explaination in the initial description and add an arg to toggle the feature. 

image

ColdShadow80:
Not much to say.
**It works, that is it. **

Request: add a placeholder to type of nearby refence (Gym/pokestop) to be sent as an additional field in the webhook  if possible.

ccev:
@ColdShadow80 You mean a field like `"type": "pokestop"`? I used to think nearby mons are sent for both gyms and stops, but as it turns out, they're just sent at Stops.

ColdShadow80:
> @ColdShadow80 You mean a field like `"type": "pokestop"`? I used to think nearby mons are sent for both gyms and stops, but as it turns out, they're just sent at Stops.

Something like that. I assumed both could be possible since your code covers that possibility, did not noticed if the posts I had were only stops or not. Assuming both can exist (and a bit of "future proof" ) a `"nearby_type": "pokestop" makes sense for me.

ccev:
More changes:

If there's no Stops in the Level 15 cells that are within 200m of an account's location, the game will send all mons in that cell as nearby. This PR now handles these correctly. I've also added a new column to the pokemon table to make it easier to understand what kind of mon it is. Additionally, I added more webhook types.

New DB columns:
- `seen_type`: An enum. Either `wild`, `encounter`, `nearby_stop`, `nearby_cell`, `lure_wild` or `lure_encounter`. (not that i'm waiting for a PD update to support lure mons)
- `cell_id` the S2 cell's ID a mon is in (if last_seen = nearby_cell)

New webhook fields:
- `seen_type` in all pokemon payloads
- `cell_id` and `cell_coords` in nearby_cell pokemon webhooks. cell_coords can be used to draw the cell as a polygon.

Webhook customization:
- You can now specify the type of mon to be sent via webhook. Like this: `[wild encounter nearby_stop nearby_cell lure_encounter]https://url.com`
- For backwards compatibility `pokemon` is an alias for `encounter`
- This will remove the args for noniv/nearby webhooks (not as of writing this comment, but it will)
- This will not break poraclePvpHelper (I made sure). The plugin will keep sending `encounter` mons. Everything else has to be configured in the MAD config
- Note that this has no effect on how Receivers handle the webhooks. The webhook-type will stay `pokemon` for all of them, just the new `seen_type` will be different

ChaosEmpire:
PvPporaclehelper webhook sends pokemon incl pvp data to http://localhost:xyz

MAD webhook send duplicate with IV even its set in config as
[wild nearby_stop nearby_cell lure_encounter lure_wild]http://localhost:xyz

ccev:
I added a few more things to make it easier to scan for nearby_cell mons:

- New db column: trs_s2cells.has_pokemon: Indicating if that cell is sending Pokemon of type nearby_cell
- When enabling Cellupdates and clicking on a cell, this value will show in the popup
- New setting for mon_mitm areas: `nearby_mon_mode` a True/False switch to optimize the area to scan for nearby_cell Pokemon. Default is False and would calculate the route we're used to. If set to True, it will only cover cells with has_pokemon = 1

The screenshot shows a route generated with this new mode. Note that I set the visual mon range to 180m in the map settings, which is also the range I use in the calculation.

image

Currently, init doesn't work well together with this mode. I'll try to get that fixed.

rfbueno:
There is any way to set pokemon spawnpoint_id around the point of pokestop? Not the exact point of pokestop, like a circle around the stop (which mon with different location) thanks!

dkmur:
> There is any way to set pokemon spawnpoint_id around the point of pokestop? Not the exact point of pokestop, like a circle around the stop (which mon with different location) thanks!

<https://discord.com/channels/465247740553592832/562515411149651983/836582132465795122>

fosJoddie:
Alright, lemme try and summarize my thoughts on this
new columns in pokemon table means potential breaking front-ends which at the very least there should be a note somewhere about how (if at all) RDM handles the same functionalities
the new column name is `seen_type` why is it not simply `source` or similar. It isn't the seen type of the pokemon, it's the source of the data. I get that seen is chosen because it has to do with how it was discovered by the scanner, but it doesn't really sit well for me
There's also columns for `fort_id` + `cell_id` in pokemon table now, not sure how I feel about that.
The new stuff in webhooks should also be aligned with whatever RDM is doing. I see no mention of this in the PR
Finally, this needs to be rebased against master, and patches squashed as Spezus mentions. Currently I'm seeing quite a few "merge with master" commits that should be removed (by interactive git rebase)

pki0:
> There's also columns for `fort_id` + `cell_id` in pokemon table now, not sure how I feel about that.
I agree totally with that.

In general MAD is build off the old RM database style. Even if the game uses 'fort_id', MAD is using 'pokestop_id' and 'gym_id'. Using other names for the same thing will be a bad mix.

Many notification bots for telegram/discord and also frontends already work with this style. Everything added later in MAD are new tables with the prefix 'trs_*'. I think we also need here a new table only with 'Encounter_id' to join the Pokemon table with the new rows. Everything else means a lot of rework and adjusting for existing tools. For example: 'trs_nearby'. So no tool will ever break by updating MAD.

The next thing is that - before it can get merged - we need PRs for all tools MAD supports in the documentation: RocketMAD, PMSF, Porcale and so on. (I don't know what tool already has the support. I only see that there is no PR for RocketMAD and it will break if this gets merged)

For now that are my thoughts for the database-thing. There are more things to do later.

jfberry:
I am ru no

—
James Berry, m. 07768 021031


________________________________

jfberry:
I am running this with RocketMAD and Poracle both working perfectly well

ccev:
Adding columns to existing tables should not be a problem for other tools. It's not the first time that's done.

ChaosEmpire:
RocketMAD does not support yet displaying the type where the pokemon was seen (helpful for the dispalyed times)
and RocketMAD is not able to change the position of a nearby seen pokemon, which got later encountered, and that way found its real spawnpoint

the basic map of mad is able to do this already, so the code "just" needs a copy/paste

webhooks get send for all corrections, and most external tools work with the fields given there and not whats in DB

so i dont see a big influence for front end

Micha's IVscript needed a changed query, to display the nearby pokemon, but it worked also fine, without a change, even it s working directly with the DB

static/madmin/static/js/madmin.js
```diff
@@ -1535,6 +1535,23 @@ new Vue({
                 var ivcolor = "red";
             }
 
+            // I have no idea how JS works and i'm sure the below block can be done better
```
Grennith, line 1538: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Statements/switch

mapadroid/worker/WorkerMITM.py
```diff
@@ -191,7 +191,16 @@ def _check_for_data_content(self, latest_data, proto_to_wait_for: ProtoIdentifie
         if latest_proto_data is None:
             return LatestReceivedType.UNDEFINED, data_found
         latest_proto = latest_proto_data.get("payload")
-        key_to_check: str = "wild_pokemon" if mode in ["mon_mitm", "iv_mitm"] else "forts"
+
+        if mode in ["mon_mitm", "iv_mitm"]:
```
Grennith, line 195: Could use enums like in WorkerFactory https://github.com/Map-A-Droid/MAD/blob/90e5f5b52f546b36f48ef18446125afa298c6c17/mapadroid/worker/WorkerFactory.py#L184

mapadroid/worker/WorkerMITM.py
```diff
@@ -191,7 +191,16 @@ def _check_for_data_content(self, latest_data, proto_to_wait_for: ProtoIdentifie
...
+            settings = self._mapping_manager.routemanager_get_settings(self._routemanager_name)
+            if settings.get("nearby_cell_mode", False):
+                key_to_check = "nearby_pokemon"
```
Grennith, line 198: Introduce enum / constants for proto keys

mapadroid/webhook/webhookworker.py
```diff
@@ -479,6 +481,23 @@ def __prepare_mon_data(self, mon_data):
                 if self.__args.quest_webhook_flavor == "poracle":
                     mon_payload["weather"] = mon["weather_boosted_condition"]
 
+            if mon["seen_type"] in ("nearby_stop", "lure_wild", "lure_encounter"):
```
Grennith, line 484: Enum for the "seen_type" would make sense imho

mapadroid/webhook/webhookworker.py
```diff
@@ -408,20 +415,15 @@ def __prepare_mon_data(self, mon_data):
...
 
-            if not self.__args.pokemon_webhook_nonivs \
-               and mon["pokemon_id"] in self.__IV_MON \
-               and (mon["individual_attack"] is None):
-                # skipping this mon since IV has not been scanned yet
-                continue
-
             mon_payload = {
-                "encounter_id": mon["encounter_id"],
+                "encounter_id": str(mon["encounter_id"]),
```
Grennith, line 419: What's the reason to make this a string?

mapadroid/mitm_receiver/MITMReceiver.py
```diff
@@ -312,9 +312,15 @@ def get_latest(self, origin, data):
         if unquest_stops is not None:
             unquest_stops = unquest_stops.get("values", [])
 
+        if self.__application_args.scan_lured_mons:
```
Grennith, line 315: check_lured = self.__application_args.scan_lured_mons

mapadroid/db/DbPogoProtoSubmit.py
```diff
@@ -78,30 +89,156 @@ def mons(self, origin: str, timestamp: float, map_proto: dict, mitm_mapper):
...
+                form = display["form_value"]
+                costume = display["costume_value"]
+                gender = display["gender_value"]
+
+                now = datetime.utcfromtimestamp(time.time())
+                disappear_time = now + timedelta(minutes=15)  # TODO: Possible config option?
+                disappear_time = disappear_time.strftime("%Y-%m-%d %H:%M:%S")
+                now = now.strftime("%Y-%m-%d %H:%M:%S")
+
+                if stopid == "" and cellid is not None:
```
Grennith, line 177: if not stopid and cellid
should work here

mapadroid/db/DbPogoProtoSubmit.py
```diff
@@ -78,30 +89,156 @@ def mons(self, origin: str, timestamp: float, map_proto: dict, mitm_mapper):
...
+                    seen_type = "nearby_cell"
+                    cell_encounters.append((encounter_id, now))
+                else:
+                    db_cell = None
+                    seen_type = "nearby_stop"
+                    stop = self._db_exec.execute(stop_query, (stopid))
+                    if (not stop) or (not len(stop) > 0) or (not stop[0][0]):
+                        stop = self._db_exec.execute(gym_query, (stopid))
+
+                    if len(stop) > 0:
```
Grennith, line 190: if stop:

mapadroid/webhook/webhookworker.py
ccev: Jabes said JS has issues working with encounter IDs since their too big. (according to him that also means it's invalid json)

rdm also sends them as Str and so does spezus' pvp plugin. It doesn't break PA or Poracle.

jfberry: My request - this number is too large for this to be valid json. This causes parsing problems in javascript - JSON.parse loses precision resulting in cut-off id numbers in poracle logs confusing users.
RDM has always been a string so precedent, and confirmed with other tool vendors no problem (and real world testing with this PR shows no issue)

Grennith: Understood, thanks

mapadroid/worker/WorkerMITM.py
ccev: Would prefer to keep this out. The `mode` variable is only used in this line, so converting it first to an enum seems like unnecessary load to me.

https://github.com/Map-A-Droid/MAD/blob/master/mapadroid/utils/MappingManager.py#L247-L249

it gets that variable from this function, which is referenced a lot by other modules that are also using strings instead of enums. Is there something I'm missing?

mapadroid/worker/WorkerMITM.py
ccev: Unsure what you mean here. `key_to_check` is set to str since it's later being passed to https://github.com/Map-A-Droid/MAD/blob/master/mapadroid/worker/MITMBase.py#L443 which only accepts str.

mapadroid/webhook/webhookworker.py
ccev: The way I'd do it:
1. Define the enum in `utils/madGlobals.py`
2. Convert selected values to enum in the DbWebhookReder.get_mon_changed_since method
3. Introduce a new key to the data returned by that method, which is then used in those if-statements

that's good?

mapadroid/worker/WorkerMITM.py
Grennith: WorkerType.IV_MITM.value evaluates to "iv_mitm" ;)

mapadroid/worker/WorkerMITM.py
Grennith: both "nearby_cell_mode" and "nearby_pokemon" could be extracted to a constant in e.g. madGlobals.py. Just so that the keys could easily be changed in a single place rather than a potential high number.
I am aware that it is not the case way too often in current master's code... Hoping to reduce all the string/dict accesses on my orm_asyncio branch

mapadroid/webhook/webhookworker.py
Grennith: Would work for me I guess. Would probably just work with enum's .value access for now as well

PR #1127, opened 2021-03-30
Fix wizard

Expl0dingBanana:
* Update wizard to pull from APKMirror
 * tox tests for wiz-rad


fosJoddie (approved):
Tested to delete pogo/rgc/pd and hitting wizzard button -> everything got downloaded

Tested to just hit wizzard on cold boot, nothing broke and nothing redownloaded

Tested to hit wizzard again while PD was downloading but after I just downloaded pogo -> everything works as expected, and it used cached APK-mirror data

PR #1129, opened 2021-04-05
Delete items by clicking - instead of holding +

crhbetz:
Niantic recently made the item count on item deletion wrap around, so clicking `-` once will make it so all items are deleted. Therefore (repeatedly) holding `+` is no longer necessary and less efficient.

The setting `inventory_clear_item_amount_tap_duration` was used for this only and thus can be removed, requiring a database patch using version number 43.

I also update the ocr test script accordingly to the changes.

Tested this over one night with tx9s and x64mini boxes running 64bit ROMs, scanning about 3200 quests. I didn't see any issues, but also no significant performance increase.


mapadroid/utils/resolution.py
```diff
@@ -66,24 +66,20 @@ def get_delete_quest_coords(self):
             click_x = int(self._screen_x) / 1.07
         return click_x + self._x_offset, click_y + self._y_offset
 
-    def get_swipe_item_amount(self):
+    def get_click_item_minus(self):
```
ccev, line 69: `click_x = int(self._screen_x) / 3.7` could come before the if/else blocks to avoid code-duplication

mapadroid/db/madmin_conversion.py
```diff
@@ -81,7 +81,6 @@
         `post_pogo_start_delay` float DEFAULT NULL,
         `restart_pogo` int(11) DEFAULT NULL,
         `inventory_clear_rounds` int(11) DEFAULT NULL,
-        `inventory_clear_item_amount_tap_duration` int(11) DEFAULT NULL,
```
ccev, line 84: I'd assume adjusting `scripts/SQL/rocketmap.sql` would also be needed

ccev (commented):
2 things I noticed, everything else looks fine to me :)

mapadroid/db/madmin_conversion.py
crhbetz: Also done. Thanks 👍

PR #1136, opened 2021-05-03
Add buddy level quest

MonsterMap:


fosJoddie:
This works fine for English, but will break for German and French locale. 
At minimum the locale files (.po*) must be updated so the indexes match again if you are up for it (readme in locale folder)

PR #1148, opened 2021-05-13
Check walker max device settings before assigning areas

ccev:
Right now, MAD will assign an area to a device, check if that area has exceeded the configured device limit and then abort the connection to that device. This caused my setup (screenshot below) to take like 30 minutes for every device to get assigned to a mon area.

This PR will not fully fix this issue, but will improve the startup time. The area assignment only takes a few minutes now.

My walker looks like this and is assigned to about 55 devices.


ccev:
Ignore me, this PR doesn't do anything 😔 didn't test enough

PR #1167, opened 2021-06-27
Fix `Win a level X or higher raid` quest

JabLuszko:
Looks like Niantic finally added megas to that.
```
'goal': {
				'condition': [{
					'type': 7,
					'with_raid_level': {
						'raid_level': [3, 4, 5, 6]
					}
				}, {
					'type': 6,
					'with_win_raid_status': {}
				}],
				'target': 1
			},
```

### DIFF
--- a/mapadroid/utils/questGen.py
+++ b/mapadroid/utils/questGen.py
@@ -210,9 +210,9 @@ def questtask(typeid, condition, target, quest_template):
     elif typeid == 8:
         if re.search(r'"type": 6', condition) is not None:
             text = _("Win {0} Raids")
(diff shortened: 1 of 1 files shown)


JabLuszko:
Niantic fixed something. j/k j/k. They fixed it Kabuto quest, but still broken in newly introduced Regigas candy one
```
'condition': [{
	'type': 7,
	'with_raid_level': {
		'raid_level': [3, 4, 5]
	}
}, {
	'type': 6,
	'with_win_raid_status': {}
}]
```

I guess wildcard here will make more sense. From what I recall @fosJoddie wanted to re-write that part, so just a hotfix I assume?

fosJoddie:
I'll add this until I can sort my own  branch in.. 👍

PR #1165, opened 2021-06-24
Don't connect to db for empty transactions

wraythex:
As suggested by Gren on discord, 'if not args: return none'
This prevents creating a connection to the database when the args variable is empty.
This should save some system resources by stopping unnecessary connections to the db.
Thanks to comstud for tracking this down and suggesting a fix.

PR #1169, opened 2021-06-30
fix some random issue in patch 23

ccev:
if an exception occurs, `existing_data` is referenced before assignment


ccev:
checks are failing because of something that's fixed in #1165 (also, merge that PR while you're at it anyway)

ccev:
https://discord.com/channels/465247740553592832/562515411149651983/859593483898454016

PR #1171, opened 2021-07-01
Fix dropped prioq coord for iv_mitm

comstud:
When multiple workers are working on the same iv_mitm area, it
appears that prioq coords can be passed between workers, but
they are never looked for:

```
    elif self._routepool[origin].has_prio_event and self.mode != 'iv_mitm':
          prioevent = self._routepool[origin].prio_coords
          route_logger.info('getting a nearby prio event {}', prioevent)
          [...]
```

Additionally, it appears some sort of stall of the route manager
can occur when coords are passed, but I am not quite sure the cause
of that.

So, I suppose either that 'iv_mitm' check above could be removed, or we
could just not attempt to pass coords to another worker when we're in iv_mitm
mode. I'm choosing the latter, because the above 'elif' makes me believe
it never should have passed one in the first place.

Also: Add some safety to ensure we enter the block of code for checking
the prioq when mode is 'iv_mitm'... and recursing when we find we have
raced with another worker and nothing is in queue.

### DIFF
--- a/mapadroid/route/RouteManagerBase.py
+++ b/mapadroid/route/RouteManagerBase.py
@@ -573,9 +573,9 @@ class RouteManagerBase(ABC):
 
             # determine whether we move to the next location or the prio queue top's item
             if self.delay_after_timestamp_prio is not None \
-               and ((not self._last_round_prio.get(origin, False) or self.starve_route) and
-                    self._prio_queue and len(self._prio_queue) > 0 and
-                    self._prio_queue[0][0] < time.time()):
+                    and (self.mode == 'iv_mitm' or self.starve_route or not self._last_round_prio.get(origin, False)) \
+                    and self._prio_queue \
+                    and (self.mode == 'iv_mitm' or self._prio_queue[0][0] < time.time()):
                 next_prio = heapq.heappop(self._prio_queue)
                 next_timestamp = next_prio[0]
                 next_coord = next_prio[1]
@@ -594,9 +594,8 @@ class RouteManagerBase(ABC):
                                              "sure you run enough workers or reduce the size of the area! (event was "
                                              "scheduled for {})", next_readable_time)
                         return self.get_next_location(origin)
-                if self._other_worker_closer_to_prioq(next_coord, origin):
-                    self._last_round_prio[origin] = True
-                    self._positiontyp[origin] = 1
+                # Passed prioq coords are never looked at by 'iv_mitm' per above.
+                if self.mode != 'iv_mitm' and self._other_worker_closer_to_prioq(next_coord, origin):
                     route_logger.info("Prio event scheduled for {} passed to a closer worker.", next_readable_time)
                     # Let's recurse and find another location
                     return self.get_next_location(origin)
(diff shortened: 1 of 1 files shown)


mapadroid/route/RouteManagerBase.py
```diff
@@ -751,8 +756,7 @@ def _other_worker_closer_to_prioq(self, prioqcoord, origin):
         temp_distance = distance_worker
 
         for worker in self._routepool.keys():
-            if worker == origin or self._routepool[worker].has_prio_event \
-                    or self._routepool[origin].last_round_prio_event:
```
comstud, line 755: I am not sure why this check exists in this loop. Worst case it should be near the top of the method.. because.. why do all of the work in this method when it's False? :)

Instead of moving it, I've chosen to just remove it. I don't see what it buys... it can only possibly be true here for 'starve_route'.

JabLuszko:
Don't have time to test or read it properly, but I see a lot of 'iv_mitm' there :)
There is also PrioQ with raids_mitm and mon_mitms and you can have starve_route set to False for those - it will do one round of route and one round of PrioQ - balancing, let's say. You sure those changes don't break anything regarding that?

comstud:
> Don't have time to test or read it properly, but I see a lot of 'iv_mitm' there :)
> There is also PrioQ with raids_mitm and mon_mitms and you can have starve_route set to False for those - it will do one round of route and one round of PrioQ - balancing, let's say. You sure those changes don't break anything regarding that?

Yes, this only affects 'iv_mitm'. The problems being fixed here are only for 'iv_mitm'... hence the checks being added for it. The problem is the missing checks! It should make sense when you have the time to read it.

Really... all of this needs to be refactored such that there's no checking of 'mode' anywhere in the base class..  Mode specific stuff should obviously be in the mode specific classes! However, I'm just trying to get a quick fix out for this out first.

mapadroid/route/RouteManagerBase.py
```diff
@@ -573,9 +573,9 @@ def get_next_location(self, origin: str) -> Optional[Location]:
 
             # determine whether we move to the next location or the prio queue top's item
             if self.delay_after_timestamp_prio is not None \
-               and ((not self._last_round_prio.get(origin, False) or self.starve_route) and
-                    self._prio_queue and len(self._prio_queue) > 0 and
-                    self._prio_queue[0][0] < time.time()):
+                    and (self.mode == 'iv_mitm' or self.starve_route or not self._last_round_prio.get(origin, False)) \
+                    and self._prio_queue \
+                    and (self.mode == 'iv_mitm' or self._prio_queue[0][0] < time.time()):
```
comstud, line 578: These changes are not strictly necessary as long as you have 'start_route' turned on for iv_mitm. However, it feels safer to have that first check there.

As for the 2nd change here... When we're in 'iv_mitm', the first tuple element is not a timestamp.. it's basically the index at which the mon exists in the iv list. While no one would ever have a list that is as long as current epoch, it is pointless to compare it against epoch... It's always going to be true for iv_mitm.

comstud:
I'm closing this in favor of: https://github.com/Map-A-Droid/MAD/pull/1177

PR #1186, opened 2021-08-01
Enable encounters in quest mode

crhbetz:
Due to PD being able to hide Pokemon on the game map, it is no longer a problem that Pokemon might be clicked instead of the Pokestop during quest scan.

Thus, checking for these clicks based on received encounter data is no longer necessary and encounters can be received without
disturbing the quest scan flow.

This PR enables the relevant settings `mon_ids_iv` and `all_mons` for pokestop areas (database change!) and changes the quest scan workflow to ignore encounter data while trying to open a pokestop.


domizei385:
Tested and working 👍

ChaosEmpire:
running sinced weeks without issues

PR #1187, opened 2021-08-01
make iv_mitm encounter like mon_mitm

crhbetz:
With UIV enabled in PD, it is no longer necessary that an `iv_mitm` worker only encounters its scheduled specific pokemon at the target location. Instead, it can jump to the locations and just encounter every mon there. This is accomplished by telling PD
it's in mons mode while MAD still schedules the `iv_mitm` area as usual.

Using a `iv_mitm` optimized IV List as usual, but with `all_mons` enabled, will still have the worker teleport to desirable mons on priority, but encounter every mon at the location.

### DIFF
--- a/mapadroid/worker/WorkerMITM.py
+++ b/mapadroid/worker/WorkerMITM.py
@@ -114,7 +114,7 @@ class WorkerMITM(MITMBase):
             # worker has to sleep, just empty out the settings...
             ids_iv = []
             scanmode = "nothing"
-        elif routemanager_mode == "mon_mitm":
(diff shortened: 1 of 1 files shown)


ChaosEmpire:
will this cluster with other iv mitm events in reach?
if so, maybe maltes "old" code to search nearby pokemon could now get more usable

crhbetz:
It does nothing but tell PD to encounter every mon it sees (that's in the IV list) instead of explicitly one.

domizei385:
👍

ChaosEmpire:
works great, please merge

PR #1207, opened 2021-09-24
Move stats saving before cache check

JabLuszko:
People complain that they liked previous behavior more.


dkmur:
been running this since created, surely ready for merge 😇

PR #1209, opened 2021-10-05
Fix questGen a little, self too many.

JabLuszko:
However no idea if that won't fail later if for example github is dead.

// I assume Joddie tested that, but looks like not :P


fosJoddie:
Oops, I didn't double check if anyone else did this before I merged the duped PR..  closing because duped

Issue #1212, opened 2021-10-08
Wizard: AttributeError: 'dict' object has no attribute 'versions'


Micha854:
I use ubuntu 20.04 with python 3.8 and get the following error when trying to update the madmin packedes:

```[10-08 02:29:14.10] [           system] [C] An unhanded exception occurred!
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 890, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
> File "start.py", line 63, in run_thread
    run_thread_old(*args, **kwargs)
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/mapadroid/MAD/mapadroid/mad_apk/wizard.py", line 97, in apk_nonblocking_download
    self.download_pogo(arch)
  File "/home/mapadroid/MAD/mapadroid/mad_apk/wizard.py", line 135, in download_pogo
    latest_pogo_info = self.find_latest_pogo(architecture)
  File "/home/mapadroid/MAD/mapadroid/mad_apk/wizard.py", line 278, in find_latest_pogo
    latest_supported = APKWizard.get_latest_supported(architecture, available_versions, self.api_token)
  File "/home/mapadroid/MAD/mapadroid/mad_apk/wizard.py", line 330, in get_latest_supported
    for ver, release in available_versions.versions.items():
AttributeError: 'dict' object has no attribute 'versions'
[10-08 02:29:14.10] [           system] [C] An unhanded exception occurred!
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 890, in _bootstrap
    self._bootstrap_inner()
    │    └ <function Thread._bootstrap_inner at 0x7fe250fff430>
    └ <Thread(Thread-113, started daemon 140606869194496)>
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
    │    └ <function install_thread_excepthook.<locals>.run_thread at 0x7fe2380f0310>
    └ <Thread(Thread-113, started daemon 140606869194496)>

> File "start.py", line 63, in run_thread
    run_thread_old(*args, **kwargs)
    │               │       └ {}
    │               └ (<Thread(Thread-113, started daemon 140606869194496)>,)
    └ <function Thread.run at 0x7fe250fff160>

[...]

PR #1213, opened 2021-10-08
Package search failure handling

Expl0dingBanana:
Fix for #1212 

Handle the issue more gracefully to prevent an UHE from occurring


Expl0dingBanana:
I have recieved confirmation this prevents the UHE from occurring and only displays the warning message.

PR #1216, opened 2021-11-01
Scroll down in google consent screen

domizei385:
...only used and tested with x96 mini, so may not fit all sizes.

### DIFF
--- a/mapadroid/ocr/screenPath.py
+++ b/mapadroid/ocr/screenPath.py
@@ -454,7 +454,8 @@ class WordToScreenMatching(object):
 
         self._logger.info("_handle_ggl_consent_screen: Click accept button")
         self._nextscreen = ScreenType.UNDEFINED
-        self._communicator.click(520, 1185)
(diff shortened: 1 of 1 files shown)


acocalypso:
This PR fixed my issue with the annoying screen.
It's now running since 3 days and I can verify that the consent screen is handled.

ccev:
Instead of tapping and holding, wouldn't it be possible to press the `Page Down` button?

crhbetz:
End button even :-) Add this or similar to `websocket/communicator.py`:
```
def end_button(self) -> bool:
    return self.__run_and_ok("touch keyevent 123", self.__command_timeout)
```

(I started looking into a OCR based solution for this. However, for some reason tesseract will find every single word of the consent window except the `Accept` 🤷 . Currently looking for `Cancel` and tapping the opposite position as `width minus x-coordinate of cancel` as a test).

acocalypso:
Merge?

cyphercris:
This pr Fix my problem.

PR #1217, opened 2021-11-03
Fix 'is_in_battle' status from gyms and add it to webhook requests

Lecnets:
Currently all 'is_in_battle' values from gyms are set to false, even when being attacked


mpw1337:
Using this since 1 week without any problems.

PR #1220, opened 2021-11-28
Surface Werkzeug and other internal logging to loguru at orig level

Artanicus:
- With the previous logic all standard logging logs are downgraded to
DEBUG5 regardless of their initial level. This will hide any fatal
errors making debugging harder.

- Since Werkzeug is quite chatty at INFO level, anything at INFO or
  below is still downgraded to DEBUG5. This maintains the current
  behavior for those log messages.


Artanicus:
Please provide feedback / approval / rejection so I can proceed accordingly.

PR #1221, opened 2021-12-18
curl, jq, default-mysql-client, cron for automation of maintenance ta…

Breee:
curl, jq, default-mysql-client, cron for automation of maintenance tasks.
Another pull request with the possibility to enable automated updates via API calls follows soon.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,11 @@ RUN export DEBIAN_FRONTEND=noninteractive && apt-get update \
 && apt-get install -y --no-install-recommends \
 build-essential \
 default-libmysqlclient-dev \
+# utility tools for maintenance tasks
+default-mysql-client \
+jq \
+curl \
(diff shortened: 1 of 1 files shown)


Artanicus:
Not sure if making the image fatter is the right direction. The "docker way" is usually to run sidecar tasks or other containers to perform other things, instead of running tangential tasks directly within the main container.

Breee:
Yeah, we can do that as well. To me it doesn't matter, as long as i can run my CI/CD pipelines to build and rollout MAD. 

Writing a sidecar container-image that runs tasks should be trivial and also works with K8s.
Something along the lines of: 
- A base image with: Alpine Linux, Cron, Python3, mysql client 
- Start the container in the same namespace/network and Mount DB scripts, mount cronjobs / scripts and probably pass the MAD-config or set ENV variables.

PR #1223, opened 2022-01-07
Update Dockerfile for py3.9

Artanicus:
Easier to test the 3.9 changes if we build the Docker image against it as well.


Artanicus:
@fosJoddie any chance you could merge this? GitHub is now building images for the branch with python3.7 instead of 3.9.

PR #1224, opened 2022-02-02
Pin cachetools version to something before 5.0

JabLuszko:
5.0 is supposedly crashing MAD and with Grenn's rework there is no point in fixing that, I guess :P


Expl0dingBanana:
Running with cachetools>=5 works for me on Python 3.8.10. I think we need to get some more information prior to restricting cachetools.

PR #1225, opened 2022-02-12
Update current weather in gym properties

Pontiky:
The aim of this PR is to get the current weather of the S2 cell where the gym is located to update its weather property.

This change allows webhook tools such PokeAlarm to finally receive the weather ID of the gym when a raid is starting to determine if the pokemon is boosted or not.

### DIFF
--- a/mapadroid/worker/WorkerQuests.py
+++ b/mapadroid/worker/WorkerQuests.py
@@ -331,7 +331,8 @@ class WorkerQuests(MITMBase):
                      'Raid', 'Teil',
                      'Élément', 'mystérieux', 'Mysterious', 'Component', 'Mysteriöses', 'Remote', 'Fern',
                      'Fern-Raid-Pass', 'Pass', 'Passe', 'distance', 'Remote Raid', 'Remote Pass',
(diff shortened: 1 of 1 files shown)


jfberry:
Poracle remembers the weather rather than needing an extra query. Is this an option?

Pontiky:
> Poracle remembers the weather rather than needing an extra query. Is this an option?

PokeAlarm could also use cached informations like Poracle soon, mostly to determine the weather. Therefore, this change wouldn't be necessary for PA and Poracle side but as it fixes the weather not updating in gym properties and not sending to webhooks, it's still good to implement.

fosJoddie (changes requested):
I see 2 "better" ways of dealing with this

1. In DbWebhookReader, do not read out weather from `weather_boosted_condition` but do the s2cell->weather table lookup directly there
2. Add a cross table for gym->weather cell and use that to join in the right weather directly in sql like so:

new table: gym_weather_cell
columns: gym_id, weather_cell_id

gym.gym_id -> gym_weather_cell.gym_id
gym_weather_cell.weather_cell_id

gym_weather_cell can then be populated with the connection to the right weaather, instead of gyms "sharing" data from weather table arbitary.

Solution # 1 is the fastest, but # 2 will be the best scaling wise

PR #1228, opened 2022-02-24
Find and read personal commands recursively, allowing dirs

Artanicus:
This simplifies organisation and for example adding multiple upstream sources of commands as different directories or submodules.

Fully backwards compatible, assuming people aren't storing commands they don't want under a subdir of personal_commands.


Artanicus:
Friendly ping, anyone have an opportunity to review this?

Artanicus:
FYI, build failure is not due to this PR:
```
 flake8...................................................................Failed
  - hook id: flake8
  - exit code: 1
  
  mapadroid/ocr/pogoWindows.py:588:47: B023 Function definition does not bind loop variable 'thresh'.
  mapadroid/route/routecalc/calculate_route_quick.py:115:12: B020 Found for loop that reassigns the iterable it is iterating with each iterable value.
```

PR #1235, opened 2022-03-11
Add missing pokealarm fields in quest webhook when using poracle flavor

Pontiky:
There is currently a conflict with the webhook worker when switching from `default` to `poracle` `quest_webhook_flavor` setting.

The `poracle` quest flavor removes a lot of default fields instead of adding the ones which need to be added. This causes the webhook sent to PokeAlarm to be invalid as some mandatory fields are missing.

- Default quest webhook sent: https://github.com/Map-A-Droid/MAD/blob/543c3d7f002cdb64f2b6209d450ec18adc905a85/mapadroid/webhook/webhookworker.py#L154
- Poracle quest webhook sent: https://github.com/Map-A-Droid/MAD/blob/543c3d7f002cdb64f2b6209d450ec18adc905a85/mapadroid/webhook/webhookworker.py#L296

All the quest fields added for `poracle` flavor are compatible for both PokeAlarm and PoracleJS as they don't use the same name and so, they don't overwrite any field. For each webhook tools, the fields which are for the other tool will be ignored, that's not a problem.

I also renamed the `boosted_weather` field in mon data (only sent to PokeAlarm) because the real information in this field is the `weather` ID (like named for PoracleJS). This removes the difference between the flavors and add logic by renaming the PokeAlarm field. PokeAlarm still supports a `weather` field name. It will not break anything for its side.

### DIFF
--- a/mapadroid/webhook/webhookworker.py
+++ b/mapadroid/webhook/webhookworker.py
@@ -150,7 +150,7 @@ class WebhookWorker:
         return ret
 
     def __construct_quest_payload(self, quest):
-        if self.__args.quest_webhook_flavor == "default":
+        if self.__args.quest_webhook_flavor == "default":  # used by PokeAlarm
             return {
                 "pokestop_id": quest["pokestop_id"],
                 "latitude": quest["latitude"],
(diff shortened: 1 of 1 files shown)


jfberry:
This is a great idea, fully supportive

fosJoddie (approved):

PR #1259, opened 2022-06-11
Handle query string in XPathForwardedFor

ribbid987:
Editing and creating of settings doesn't work when enable_x_forwarded_path_madmin is set if query string is not handled


Grennith:
Thanks! :)

PR #1260, opened 2022-06-12
Fix raid PrioQ scanning wrong raids

landlord11:
Raid PrioQ was prioritizing hatched raids that were already scanned. Correct it to prioritize hatched raids that are still unknown.


Grennith:
Thanks :)

PR #1261, opened 2022-06-14
Allow devices to download packages

wraythex:
Devices couldn't download latest package.  The console gave the following error: "TypeError: object async_generator can't be used in 'await' expression"
Removing the "await" allows devices to download packages.


Grennith:
Given the DB APK storage requires the await, it is needed in the line which would be changed by this PR.
I have fixed it for file storage usage in [1e5fdd0](https://github.com/Map-A-Droid/MAD/commit/1e5fdd0ecbd84b417d6c34941f67173094a77dbc)

PR #1269, opened 2022-09-14
trs_status had lat/lon swapped

fosJoddie:
currentPos and lastPos are geometrical POINT datatype. when represented as WKT this is always lon, lat.  This swaps lat/lon so that it is saved correct in db

Funny enough, there doesn't seem to be anywhere internally in MAD that this db field is actually used. We use the in-memory objects instead.


Grennith:
Thanks joddie

PR #1271, opened 2022-09-29
Add support for experience quest reward

wraythex:
This PR:
 - Adds experience amount to database in the existing quest_item_amount column
 - Adds experience image to be shown on map (custom/locally made)
 - Adds logic to show experience on madmin map
 - Adds logic to show experience in madmin quest list


Grennith:
Thanks :)

wraythex:
I'm sorry to report, I did some extra testing and found an issue.  The quest_item_amount is a tinyint which can only hold values from -128 -> +127.  Since the experience reward was 700, the stored value was getting truncated down to 127 (the max of the tinyint).  I have a solution to fix this by storing the experience in the stardust column, which is a smallint which can hold up to 32767.  I'm not sure if we can revert this PR so I can fix it, or if you'd want me to create a new PR?

Grennith:
Doesn't sound like a breaking issue, feel free to PR whenever it fits your schedule :)

PR #1274, opened 2022-10-02
Fix Quest Statistics showing all stops count

JabLuszko:
However no idea why we using POLYGON here not just re-using GeofenceHelper from `fence[1]` and using min/max_lat/lon? Only place calling this with `fence` seems to be stats `statistics/GetStopQuestStatsEndpoint.py`, that already have geofence_helper and we can just do there `stops_in_fence: List[Location] = await PokestopHelper.get_locations_in_fence(self._session, geofence_helper=fence[1]` and that whole POLYGON part is not needed at all? However probably nice to have it in long run?


Boby360:
Works as intended

fosJoddie:
it (should) always be quicker to let SQL exclude by polygon than let python refine a list of roughly filtered stops.

But I agree, this whole get_locations_in_fence should be rewritten

PR #1281, opened 2022-10-03
Fix missing spawnpoints stats per geofence_id

JabLuszko:
Function expected int, got string, int != str always returned true.


Boby360:
Works as intended.
Tested with multi-area geofences as well.

PR #1284, opened 2022-10-17
reverse buggy code cellupdate code

fosJoddie:
Remove code that mistakenly was commited to only show cellupdate cells that was less than 50k (seconds/milliseconds?) old

needs a test before merge

### DIFF
--- a/static/madmin/static/js/madmin.js
+++ b/static/madmin/static/js/madmin.js
@@ -240,8 +240,7 @@ new Vue({
                     quests: 40,
                     mons: 67
                 }
-            },
-            cellUpdateTimeout: 50000
+            }
         }
     },
     computed: {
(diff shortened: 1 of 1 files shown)


static/madmin/static/js/madmin.js
```diff
@@ -908,42 +904,32 @@ new Vue({
...
             this.mapGuardedFetch("cellupdates", "get_cells" + urlFilter, function (res) {
                 const now = Math.round((new Date()).getTime() / 1000);
 
                 res.data.forEach(function (cell) {
                     const id = cell["id"];
 
-					var noSkip = true;
-                    var notTooOld = true;
-                    if (this.cellupdates[id]) {
+                    iif (this.cellupdates[id]) {
```
fosJoddie, line 913: iif -> if

PR #1285, opened 2022-10-25
Simplify setting for ✅ in simpler wording

fosJoddie:
Hide the detailed explanation behind a ?-tooltip on the status-page


fosJoddie:
Result:
image

PR #1301, opened 2023-02-12
Don't need dataclasses package anymore.

Boby360:
From what I recall, this package was only needed prior to Python 3.7.
Ran my install without the package installed with no obvious related issues.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,6 @@ asyncio-rlock==0.1.0
 bitstring==3.1.9
 cachetools==5.2.0
 ConfigArgParse==1.5.3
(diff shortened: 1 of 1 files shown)


Grennith:
We do use dataclasses in multiple instances afaik.
E.g., https://github.com/Map-A-Droid/MAD/blob/async/mapadroid/updater/SubJob.py#L1
Guessing it has been integrated into python natively since...
"New in version 3.7."
https://docs.python.org/3/library/dataclasses.html

Boby360:
That is what I meant to convey.
Yes, it is integrated natively now.

PR #1303, opened 2023-02-23
Update Redis to newer package

Boby360:
Replaces aioredis with redis, which is the successor.
Also installs Hiredis for better performance.


Boby360:
Woah, this PR was not suppose to have all of this crap in it.

PR #1310, opened 2023-03-06
FEATURE: PTC login rate limiting in redis

domizei385:
port from legacy (former master) branch

### DIFF
--- a/mapadroid/ocr/screenPath.py
+++ b/mapadroid/ocr/screenPath.py
@@ -15,6 +15,7 @@ from mapadroid.mapping_manager.MappingManagerDevicemappingKey import \
     MappingManagerDevicemappingKey
 from mapadroid.ocr.screen_type import ScreenType
 from mapadroid.utils.collections import Login_GGL, Login_PTC, ScreenCoordinates
+from mapadroid.utils.logging import get_origin_logger
 from mapadroid.utils.madGlobals import ScreenshotType, application_args
 from mapadroid.websocket.AbstractCommunicator import AbstractCommunicator
 
@@ -26,11 +27,11 @@ class LoginType(Enum):
 
 
 class WordToScreenMatching(object):
-    def __init__(self, communicator: AbstractCommunicator, pogo_win_manager, origin, resocalc,
-                 mapping_mananger: MappingManager):
+    def __init__(self, communicator: AbstractCommunicator, pogo_win_manager, origin, resocalc, mapping_manager: MappingManager):
         # TODO: Somehow prevent call from elsewhere? Raise exception and only init in WordToScreenMatching.create?
         self.origin = origin
-        self._mapping_manager = mapping_mananger
+        self._logger = get_origin_logger(logger, origin=origin)
+        self._mapping_manager = mapping_manager
         self._ratio: float = 0.0
 
         self._logintype: LoginType = LoginType.UNKNOWN
@@ -50,9 +51,9 @@ class WordToScreenMatching(object):
 
     @classmethod
     async def create(cls, communicator: AbstractCommunicator, pogo_win_manager, origin, resocalc,
-                     mapping_mananger: MappingManager):
+                     mapping_manager: MappingManager):
         self = WordToScreenMatching(communicator=communicator, pogo_win_manager=pogo_win_manager,
-                                    origin=origin, resocalc=resocalc, mapping_mananger=mapping_mananger)
+                                    origin=origin, resocalc=resocalc, mapping_manager=mapping_manager)
         self._accountindex = await self.get_devicesettings_value(MappingManagerDevicemappingKey.ACCOUNT_INDEX, 0)
         self._screenshot_y_offset = await self.get_devicesettings_value(
             MappingManagerDevicemappingKey.SCREENSHOT_Y_OFFSET, 0)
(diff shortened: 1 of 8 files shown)


mapadroid/utils/walkerArgs.py
```diff
@@ -370,6 +370,23 @@ def parse_args():
     parser.add_argument('-rrqi', '--redis_report_queue_interval', default=30, type=int,
                         help='Report queue size from mitmreciver to redis every every N seconds (Default: 30)')
 
+    # PTC Login tracking
+    parser.add_argument('-elt', '--enable_login_tracking', action='store_true', default=False,
+                        help=('Enable tracking of login attempts to PTC'))
+    parser.add_argument('-lth', '--login_tracking_host', default='localhost',
```
Grennith, line 376: Is this another redis instance rather than using buckets or unique keys not used in another part of MAD?

crhbetz: Yes, because this is a 1:1 port of a version of this I had made for master/legacy some time ago, where redis was not mandatory.

mapadroid/ocr/screenPath.py
```diff
@@ -696,3 +709,26 @@ async def _take_and_analyze_screenshot(self, delay_after=0.0, delay_before=0.0,
         else:
             returntype, global_dict, self._width, self._height, diff = result
             return returntype, global_dict, diff
+
+    async def track_ptc_login(self, mode="login"):
```
crhbetz: * `mode` is not needed anymore - the code is only checking at start of pogo, because I don't think there's any case where we try a PTC login without (re-)starting the app shortly before. So tracking at login time too would count two login attempts when only one happens.
* this function doesn't really belong into `ocr/screenPath.py`, I just slapped it in there when I came up with this idea because it was convenient at the time...

mapadroid/utils/walkerArgs.py
```diff
@@ -370,6 +370,23 @@ def parse_args():
...
+                        help=('Redis host used by login tracking'))
+    parser.add_argument('-ltp', '--login_tracking_port', default=6379,
+                        help=('Redis port used by login tracking'))
+    parser.add_argument('-ltu', '--login_tracking_username', default='',
+                        help=('Redis username for login tracking'))
+    parser.add_argument('-ltpwd', '--login_tracking_password', default='',
+                        help=('Redis password for login tracking'))
+    parser.add_argument('-ltdb', '--login_tracking_database', default=0,
+                        help=('Redis database used by login tracking. Use different numbers (0-15) if you are running multiple instances'))
```
crhbetz, line 385: Shouldn't use *different*, but the *same* numbers to use login tracking across multiple instances effectively ...

mapadroid/websocket/communicator.py
```diff
@@ -106,6 +108,34 @@ async def magisk_on(self) -> None:
     async def turn_screen_on(self) -> bool:
         return await self.__run_and_ok("more screen on\r\n", self.__command_timeout)
 
+    async def get_external_ip(self) -> Optional[str]:
+        try:
+            res = await self.passthrough("echo \"$(curl -k -s https://ifconfig.me)\"")
```
crhbetz, line 113: Probably cleaner to make the external IP provider configurable?

mapadroid/worker/strategy/AbstractWorkerStrategy.py
```diff
@@ -210,6 +212,15 @@ async def start_pogo(self) -> bool:
             await asyncio.sleep(
                 await self.get_devicesettings_value(MappingManagerDevicemappingKey.POST_TURN_SCREEN_ON_DELAY, 7))
 
+        if await self.get_devicesettings_value(MappingManagerDevicemappingKey.LOGINTYPE, 'google') == 'ptc':
+            if self._args.enable_login_tracking:
```
crhbetz: args are imported as `application_args` (lines 25 to 29), so the passing around of args/self.args across all those Strategy classes is no longer needed.

mapadroid/utils/walkerArgs.py
domizei385: Nonetheless. I use a single redis for tracking in multiple mad instances, so it would still need to be configurable

mapadroid/ocr/screenPath.py
domizei385: Right. I removed mode.
As for the placement, I am open for ideas.

mapadroid/utils/walkerArgs.py
domizei385: fixed

mapadroid/worker/strategy/AbstractWorkerStrategy.py
domizei385: Thanks. Removed.

mapadroid/websocket/communicator.py
domizei385: Not sure, if that belongs into the config.ini though?!

mapadroid/worker/strategy/AbstractWorkerStrategy.py
crhbetz: I think all changes to [mapadroid/worker/strategy/StrategyFactory.py](https://github.com/Map-A-Droid/MAD/pull/1310/files#diff-50accc4a526b66551ef7c831220f9e598b88260e94c2c4a3313f01c0f2e3151f) also relate to passing around `args` so those would all be able to be removed too.  🤔

domizei385: reverted

domizei385:
Closed, as this was integrated into async_account_swtiching

PR #1313, opened 2023-03-09
add missing incident display_type for eventstops in wh

ReuschelCGN:
for poracle eventstop tracking like kecleon and goldstops WH must contains the display type...


Grennith:
Strictly speaking the incident_display_type is added for each incident.

PR #1317, opened 2023-03-14
Update AccountHandler.py

nakratz:
mark_burnt was not updating the account once BSOD (maintenance screen) was being observed. Was missing session.commit()

Thanks to Jab for pointing out location of missing commit in discord.


Grennith:
I already pushed that when I was checking it sorta in parallel. Sorry and thanks!

PR #1319, opened 2023-03-14
Fix: Reset self_screen_count before rebooting the device

crhbetz:
Experienced some edge case where this counter would not reset after trying to reboot the device, because the worker instance was retained - so it would be possible that another reboot is triggered quicker than intended.


crhbetz:
Already fixed by https://github.com/Map-A-Droid/MAD/commit/8c29aad0e2fff94f2d9734cbd061257dc58653e0 :-)

PR #1327, opened 2023-03-19
title tooltip with last_burn time

JabLuszko:
I somehow forgot to include it in first PR, sorry. No, it's not mobile friendly probably :D

https://i.imgur.com/nknDVKu.png / https://i.imgur.com/tq7poD7.png


JabLuszko:
Moved everything to javascript (no idea if this won't kill a browser if it has to do thousands of moment.js time parsing, should not?)
Changes icons because they were confusing to some nation that of course had road signs done differently than rest of the world.

https://i.imgur.com/YQvRnWG.png

Now it's a `flag` for flagged accounts, but ready to use and `hourglass` for accounts on cooldown.

PR #1342, opened 2023-04-20
Handle PoGo tutorial via ocr

S1lverhand:
This commit handles the Pokémon GO tutorial for new (PTC) accounts via OCR. 

There are several "entry points" for the tutorial, which can be found by restarting PoGo on each screen of the tutorial. Each new ScreenType defines one of the entry points and each new function handles the screens after this entry point up to the next entry point. If an error occurs in one function (e.g., no starter pokémon can be catched), PoGo is restarted. The first screen after the restart will be an entry point and tutorial handling should continue.

Most of the credit goes to @cecpk and his fabulous OCR handling.

The commit has been tested only on tx9s with screen resolution 1080x1920 for PTC accounts. Other resolutions should also work because I used relative screen coordinates.
I don't know, if the tutorial handling works for Google accounts. Would be nice, if someone could test this.
Slower devices might need larger sleep times. I thought of introducing a general tutorial_sleep_time_multiplier, which can be set in the device settings, but I don't know how to do this.


Grennith:
Thanks a ton for taking up the OCR stuff :)

PR #1347, opened 2023-05-07
Log incidents cleanup

JabLuszko:
Moar logs, moar fun.


mpw1337 (approved):
Looks good to me, had a similar adjustment on my instance.

PR #1353, opened 2023-06-13
Use AccountHandler when getting account data, reason for switch_account/switch_user

domizei385:
Added some missing interactions with AbstractAccountHandler, added reason to switch_user/switch_account to easier work with an external accountHandler

### DIFF
--- a/mapadroid/worker/strategy/AbstractWorkerStrategy.py
+++ b/mapadroid/worker/strategy/AbstractWorkerStrategy.py
@@ -6,6 +6,7 @@ from typing import Any, List, Optional
 
 from loguru import logger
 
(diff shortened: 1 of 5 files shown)


mapadroid/worker/strategy/AbstractWorkerStrategy.py
```diff
@@ -410,7 +411,7 @@ async def _restart_pogo_safe(self):
         await asyncio.sleep(1)
         return await self.start_pogo()
 
-    async def _switch_user(self):
+    async def _switch_user(self, reason=None):
```
Grennith, line 414: Where is that reason used? What type should it have? Apparently string? Could use a type (imho enum) at this point

PR #1355, opened 2023-06-25
correct delay_to_avoid_softban

landlord11:
The current calculation adds the total delay needed to the time elapsed.  It should subtract the time elapsed since last softban from the delay needed.  For example, if distance jumped requires a 10s delay and 7s have elapsed, the old code calculated a 17s delay is needed.  It should be 3s.


Grennith:
Thanks, nice finding :)

PR #1362, opened 2023-09-01
New screen Limitations aka most likely better MAINTENANCE

JabLuszko:
Copy-paste code at this moment. If this turns out to be the old MAINTENANCE thing without any other side-effects we could streamline that in future and merge it into one, but for now copy-paste for faster debugging.

Default 24h cooldown, normal MAINTENANCE burnt procedure.

Also need different languages, but PokeMiners don't update language files anymore so DE/FR translation most welcome

### DIFF
--- a/mapadroid/ocr/screenPath.py
+++ b/mapadroid/ocr/screenPath.py
@@ -267,8 +267,6 @@ class WordToScreenMatching(object):
             if MadGlobals.application_args.enable_early_maintenance_detection and self._worker_state.maintenance_early_detection_triggered:
                 logger.warning("Seen RETRY screen after multiple proto timeouts - most likely MAINTENANCE")
                 await self._account_handler.mark_burnt(self._worker_state.device_id, BurnType.MAINTENANCE)
-            # forcing clear_game_data here due to Niantic changes and game now remembering/pre-filling username on login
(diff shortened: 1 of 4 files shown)


JabLuszko:
Niantic seems to introduced some QOL changes and after failed login/maintenance hit (RETRY screen, it still does not show up live in-game) the username field is getting pre-filled/remembered across game restarts. MAD flow does not handle that and every login attempts it adds login again to already pre-filled login

Forcing cleaning game data/cache in RETRY screen - this means MAD will need OCR birthday again... but seems better choice then sending backspace 50 times on each login attempt or even worse detecting which (correct or older?) login is pre-filled there - so we won't end up like that https://discord.com/channels/465247740553592832/523602714244939787/1147526555136966726 :D

JabLuszko:
>  it still does not show up live in-game
Okey, this is a lie. 

That screen shows up in-game so we probably could run screen OCR before `Too many timeouts - Restarting game` to see if it's MAINTENANCE and save on one game restart (to see RETRY screen)... but no idea if this is worth it

Grennith:
Merging for now as a hotfix measurement. Will align/review code to integrate it via OCR in the next couple of days.

PR #1375, opened 2024-01-05
DataTable'd pogoauth settings

JabLuszko:
50 results per page, sorted by device column desc (so accounts tied to devices first in results).
Plus as bonus we get searchbar :D

PR #1376, opened 2024-01-09
Fix #1375 DataTable'd pogoauth setting

JabLuszko:
copy-paste is hard :(


mpw1337 (approved):
Looks good, fixed it like this on my instance. Works like a charm now 👍

PR #1386, opened 2024-09-08
WIP: Stations

JabLuszko:
```
stop mad
(backup db :D)
$ git fetch origin pull/1386/head:stations_pr
$ git checkout stations_pr 
start mad
```
Added **Stations** to database (not in mapview/stats).
Fixed routes not saving due to changed proto, from my test seems like taking last element is enough - it's accepted or it's not in protos, at all.
Fixed raids not saving due to changed proto (`is_exclusive` is gone, force 0)
Added ssl=False to RestHelper for supported versions query and outdated cert there.
Added `Base` to alembic env so can generate automigration in future.

#TODO: 
stats? /do we really need stats for this and I don't really want to touch stats, especially webview/


JabLuszko:
So this seems to be working.

even more #TODO
Do we need move1/move2?
Is reward pokemon needed - most likely not, but maybe with some forms/evo in future?

HELLspawnTIM:
Many thx for that. I guess this also requires some work on map gui for the users, e.g. RocketMAD?

JabLuszko:
#TODO?

Niantic deleted some powerspots that should be still active according to `end_time_ms`. They are gone from protos, they will show up in-game if you haven't restarted client and crash it when attempting to interact (LOL).
So cleanup? -> add cell_id to db, base cleanup on last_updated not updated within X, but s2cellid updated/processed by MAD?

JabLuszko:
Added `move_1` and `move_2`, could be useful for bigger raids (like avoid Kyogre Blizzard)
Not adding `stamina`, seems very behind the scene thing.
Adding `bread_mode` as I assume this is gonna be to used to distinguish between Dyna and Giga.
Not adding `cell_id` - seems that deletion of active stations was one time only N feature. 
`webhook support` - let's wait for Poracle to support this first, I guess. We can adjust format to theirs liking
`stats/map view` - I don't want to touch it, I have no idea what is happing there. 
Added `number_stationed_pokemon` to DB but this need (another) external proto PD support, so just NULL/placeholder for now.
Removed `cooldown` thing - seems like battle_cooldown per account.

#TODO:
DB Migration

JabLuszko:
You can test this now, shouldn't explode MAD.
Instructions in first post.

JabLuszko:
~~I am not sure if not updating reward_pokemon/battle_pokemon and battle times to null makes sense.
Powerspot seen after hours (so like after 21:00) won't update this to null and it will stay as maxbattle in database. Could be nice for some historical/check view, but does not translate 1:1 to game.~~


~~However it's also nice because after that 21:00 I can take a look at map and plan a walk for tomorrow on already existing maxbattles (and maybe also get some new...). Plus this also will work for 'rarely' scan areas - if the powerspot is active for 3 days we know (at least for now) that they "same" maxbattle shows up everyday 6:00 - 21:00 and if we scan it once we can assume we have data for 3 days and just handle that in front-end.~~

Seems like old maxbattle data is still in proto after 21:00 if powerspot is valid next day. If it disappears 5:00 the maxbattle data is gone.

JabLuszko:
Added webhook sending data, just 1:1 fields in proto(names). I don't see any PR in PoracleJS so we will adjust it later I guess.

We need to bump REDIS cache key to something more than 15 minutes - there is no need to update those every 15 minutes.

JabLuszko:
There is a chance we will need to adjust redis cache key to include battle_pokemon.pokemon_id or bread_battle_seed.
Seems like mon-boss is changing mid-day mid-battle window, but I need it to first hit my timezone to double check.

I hope we can do bread_battle_seed tbh as pokemon_id could be not enough if Niantic one day decide to run (f.e.) both Ponyta's in some raid day or something - hour each - those will only differ by form_id. Or normal Pikachu and some costume one.

JabLuszko:
> There is a chance we will need to adjust redis cache key to include battle_pokemon.pokemon_id or bread_battle_seed. Seems like mon-boss is changing mid-day mid-battle window, but I need it to first hit my timezone to double check.
> 
> I hope we can do bread_battle_seed tbh as pokemon_id could be not enough if Niantic one day decide to run (f.e.) both Ponyta's in some raid day or something - hour each - those will only differ by form_id. Or normal Pikachu and some costume one.

We don't need to, but I have a feeling we still should use battle_seed.
Some powerspots that should be active for (f.e.) 2 more days with Charmander changed to Beldum - with totally new `start_time`, `end_time` and everything in `battle_` so they were detected as "new" powerstop and overwritten the one in database by `station_id` and trigger webhooks.

JabLuszko:
Adjusted database schema so it will work with reactMap. No migration, you need to drop table and set alembic_version back to 'async'.
```
UPDATE `alembic_version` SET `version_num` = "b533c33be802";
DROP TABLE `station`;
```
before using new version of this PR.

JabLuszko:
> Adjusted database schema so it will work with reactMap

Not really, they expect `int` we using `datetime` and I don't think we can adjust that as every other place we use datetime.
There will be a need for a PR to reactmap
```
diff --git a/server/src/models/Station.js b/server/src/models/Station.js
index bd2a92bb..1b1d113d 100644
--- a/server/src/models/Station.js
+++ b/server/src/models/Station.js
@@ -1,5 +1,5 @@
 // @ts-check
-const { Model } = require('objection')
+const { Model, raw } = require('objection')
 const config = require('@rm/config')
 const i18next = require('i18next')

@@ -30,24 +30,28 @@ class Station extends Model {
       'name',
       'lat',
       'lon',
-      'updated',
-      'start_time',
-      'end_time',
+      isMad ? raw('UNIX_TIMESTAMP(updated) as updated'): 'updated',
+      isMad ? raw('UNIX_TIMESTAMP(start_time) as start_time') : 'start_time',
+      isMad ? raw('UNIX_TIMESTAMP(end_time) as end_time') : 'end_time',
       'total_stationed_pokemon',
     ]

     const query = this.query()
       .whereBetween('lat', [args.minLat, args.maxLat])
       .andWhereBetween('lon', [args.minLon, args.maxLon])
-      .andWhere('end_time', '>', ts)
+      .andWhere(
+        isMad ? raw('UNIX_TIMESTAMP(end_time)') : 'end_time',
+        '>',
+        ts,
+      )
     // .where('is_inactive', false)

     if (perms.dynamax && onlyMaxBattles) {
       select.push(
         'is_battle_available',
         'battle_level',
-        'battle_start',
-        'battle_end',
+        isMad ? raw('UNIX_TIMESTAMP(battle_start) as battle_start') : 'battle_start',
+        isMad ? raw('UNIX_TIMESTAMP(battle_end) as battle_end') : 'battle_end',
         'battle_pokemon_id',
         'battle_pokemon_form',
         'battle_pokemon_costume',
@@ -218,7 +222,7 @@ class Station extends Model {
       .whereBetween('lat', [bbox.minLat, bbox.maxLat])
       .andWhereBetween('lon', [bbox.minLon, bbox.maxLon])
       .andWhere(
-        'updated',
+        isMad ? raw('UNIX_TIMESTAMP(updated)') : 'updated',
         '>',
         Date.now() / 1000 - stationUpdateLimit * 60 * 60 * 24,
       )

```

JabLuszko:
Seems like Niantic yet again spawned phantom stations (even better - for next day) and then deleted them ASAP?

JabLuszko:
and another wave of spawns of powerspots for tomorrow however not deleted today. Third time a charm? N figured out how it works? :D

PR #1387, opened 2024-09-20
Fix wizzard apkm/zip split packages.

JabLuszko:


ofrinach:
Hey @JabLuszko , 
Line 124 in `MadApkEndpoint.py` should also be changed to allow .apkm files, otherwise a "Unsupported extension" response is returned

JabLuszko:
> Hey @JabLuszko , Line 124 in `MadApkEndpoint.py` should also be changed to allow .apkm files, otherwise a "Unsupported extension" response is returned

True, totally forgot I renamed my testing .apkm to .zip.

mapadroid/madmin/endpoints/api/apks/MadApkEndpoint.py
```diff
@@ -39,6 +38,10 @@ async def get(self):
             except KeyError:
                 return await self._json_response(data=data[apk_type])
 
+    def allowed_file(self, filename):
+        ALLOWED_EXTENSIONS = set(['apk', 'apkm', 'zip'])
```
Expl0dingBanana, line 42: This line is equivalent to `ALLOWED_EXTENSIONS = {'apk', 'apkm', 'zip'}` which is more efficient as it doesnt create a list then convert it to a set. I dont think it matters in this context though.

This should be a global variable thats initialized once and referenced instead of initialized on every call.

mapadroid/mad_apk/wizard.py
```diff
@@ -537,12 +538,13 @@ def normalize_package(self) -> NoReturn:
                         apk = apkutils.APK().from_io(bytes_of_apk).parse_resource()
                         manifest = apk.get_manifest()
```
Expl0dingBanana, line 539: It looks like manifest is no longer used. Should be good to remove the line.

mapadroid/madmin/endpoints/api/apks/MadApkEndpoint.py
```diff
@@ -58,7 +61,7 @@ async def post(self):
             elif not file.filename:
                 await self._add_notice_message('No file selected for uploading')
                 raise web.HTTPFound(self._url_for("upload"))
-            elif not allowed_file(file.filename):
```
Expl0dingBanana, line 61: Is `allowed_file` utilized anywhere else? If not, the code should be removed to have a cleaner code base.

